PR #58, opened 2022-06-08
2861738 Scaffold E2E test project

tejaswip-msft:
# What
Scaffold for E2E test project

# Why

# How Tested

# Process & policy checklist


tejaswip-msft (commented):
The example test is auto generated when initializing playwright. Did not remove it as it can serve as a reference when implementing test cases

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "sample",
+  "version": "1.0.0",
+  "description": "![vv-banner.png](./docs/images/vv-call-desktop.png)",
```
tejaswip-msft, line 4: This is taken by default from README file. Shall i change it to an empty string or some other reasonable description?

tejaswip-msft (commented):
The snapshots does not have the iframe. Handling iframe in snapshots require a different approach and can be dealt with the further E2E stories

hansung-msft:
Would you be able to provide some steps for how to run / test it please?

package.json
hansung-msft: I am not sure if we want to use "description" in this `package.json`. In my opinion, I think either remove this property or put empty string make sense to me.

I would like to see what others think abou this!

allenplusplus: Agreed, I think we either omit the description property or change this to another string. The way the string is formatted now is meant to be in markdown file.

allenplusplus:
I suggest moving all the playwright and tests files into their own project folder. The current setup turns the entire repository into a single npm project with nested npm projects. It looks like this:
```
- / (has project.json)
   - client (has project.json)
   - server (has project.json)
   - tests (has no project.json)
```

I feel like this is a bit unnatural to me and couples the projects together. I would suggest a structure where each project is its own individual node project. For example:
```
- / (no project.json here)
    - client (has project.json)
    - server (has project.json)
    - e2e_tests (has project.json)
```

playwright.config.ts
```diff
@@ -0,0 +1,113 @@
...
+  projects: [
+    {
+      name: "chromium",
+      use: {
+        ...devices["Desktop Chrome"],
+      },
+    },
+
+    {
+      name: "firefox",
```
allenplusplus, line 59: UI-Library doesn't support firefox. I think we can skip testing in firefox in this case

playwright.config.ts
```diff
@@ -0,0 +1,113 @@
...
+    // baseURL: 'http://localhost:3000',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: "on-first-retry",
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: "chromium",
```
allenplusplus, line 52: Does the chromium cover Edge? If not I think we should add one for Edge browser

hansung-msft: I might be wrong but I thought Chromium is open source browser and I think Edge is based on Chromium..

hansung-msft: Oh I am not sure if that "chromium" is meant to "Chrome" browser or not..

hansung-msft:
I tried to run it.. but I got errors as the screenshot below. I must doing something wrong here 😢 .

image

hansung-msft (changes requested):
I tried to run it.. but I got errors as the screenshot below. I must doing something wrong here 😢 .

image

playwright.config.ts
```diff
@@ -0,0 +1,113 @@
...
+    //   use: {
+    //     ...devices['iPhone 12'],
+    //   },
+    // },
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: {
+    //     channel: 'msedge',
```
yilunqian-msft, line 90: Let's remove these lines instead of commenting them out.

playwright.config.ts
```diff
@@ -0,0 +1,113 @@
...
+// Licensed under the MIT license.
+
+import type { PlaywrightTestConfig } from "@playwright/test";
+import { devices } from "@playwright/test";
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
yilunqian-msft, line 11: remove

playwright.config.ts
tejaswip-msft: Okay, i did not remove as i am not sure if we want to implement these. These are autogenerated and already commented out when installed

tejaswip-msft:
> Would you be able to provide some steps for how to run / test it please?

- To install supported browsers, use the following command:
```
npx playwright install

```
- To build and run the tests, use the following commands:

```
npm run build:e2e
npm run test
```

hansung-msft:
> > Would you be able to provide some steps for how to run / test it please?
> 
> * To install supported browsers, use the following command:
> 
> ```
> npx playwright install
> ```
> 
> * To build and run the tests, use the following commands:
> 
> ```
> npm run build:e2e
> npm run test
> ```

I was able to run the test and all success after run `npx playwright install`. 
image

hansung-msft (approved):
Looks good to me for scaffolding.

hansung-msft (approved):
Just tested with Microsoft Edge added, it worked on my end 😄 

image

e2e-tests/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "sample",
```
allenplusplus, line 2: Minor one but I think we should name this like e2e-tests instead of sample

allenplusplus (approved):

PR #61, opened 2022-06-17
Update snapshot workflow

tejaswip-msft:
# What
Workflow to update snapshots whenever there is a UI change

# Why
If a PR has a UI change, a label "ui change" should be added. It will trigger this workflow which will update the snapshots required for E2E testing

# How Tested
Forked the repo and tested there
Workflow run : https://github.com/tejaswip-msft/Sample/runs/6928184675?check_suite_focus=true
Workflow added a commit to the branch with updated screenshot : https://github.com/tejaswip-msft/Sample/pull/1/commits/cab5428d1651bc0f1ef19e68a232fa90d3e6663e

# Process & policy checklist


ralphgabrinao:
Looks great!

I can't find it in the workflow, but are we also going to add the automation of this comment? Or is it going to be in a separate PR? - https://github.com/Azure/communication-ui-library/pull/1960#issuecomment-1156181864

Saying "Failed to pass the UI Test. If this PR is for UI change and the error is snapshot mismatch, please add "ui change" label to the PR for updating the snapshot."

tejaswip-msft:
> Looks great!
> 
> I can't find it in the workflow, but are we also going to add the automation of this comment? Or is it going to be in a separate PR? - [Azure/communication-ui-library#1960 (comment)](https://github.com/Azure/communication-ui-library/pull/1960#issuecomment-1156181864)
> 
> Saying "Failed to pass the UI Test. If this PR is for UI change and the error is snapshot mismatch, please add "ui change" label to the PR for updating the snapshot."

Yes this will be added in a separate PR

e2e-tests/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```
+npx playwright install
+
+```
+
+- To build and run the tests, use the following commands:
+
+```
+cd e2e-tests
```
ralphgabrinao, line 17: Since we are already inside the e2e-tests folder, we don't need this anymore.

e2e-tests/playwright.config.ts
```diff
@@ -48,21 +48,6 @@ const config: PlaywrightTestConfig = {
...
-      },
-    },
-
-    /* Test against branded browsers. */
-    {
-      name: "Microsoft Edge",
-      use: {
-        channel: "msedge",
-      },
-    },
```
ralphgabrinao, line 65: Curious, why were these removed?

tejaswip-msft: So these browsers are not supported while running in linux machine, locally they pass, but in pipeline they fail :(

ralphgabrinao: Ah, they are platform dependent as well. I see UI lib tests on Chromium only also, so I think this is fine.

ralphgabrinao (approved):

.gitignore
```diff
@@ -88,6 +88,11 @@ docGen/
...
 # Release outputs
 release/
+
+# e2e tests outputs
 e2e-tests/test-results/
 e2e-tests/playwright-report/
 e2e-tests/playwright/.cache/
+
+# Snapshots generated while running e2e tests locally
+*win*.png
```
tejaswip-msft, line 98: we dont want locally generated snapshots in the repo, the pipeline will generate linux snapshots and the tests will be validated against them

ralphgabrinao: I think here we can be more specific with the path, so as to not exclude **all** filenames that match that pattern : `e2e-tests/tests/snapshots/**/*win*/png`

.gitignore
```diff
@@ -88,6 +88,11 @@ docGen/
 
 # Release outputs
 release/
+
+# e2e tests outputs
 e2e-tests/test-results/
 e2e-tests/playwright-report/
 e2e-tests/playwright/.cache/
+
+# Snapshots generated while running e2e tests locally
```
ralphgabrinao, line 97: Let's also expound on why we don't want to commit them. Could be something along these lines:

`Screenshots differ between browsers and platforms due to different rendering, fonts and more. Your Windows development environment may differ from someone else's. For more information, visit https://playwright.dev/docs/test-snapshots`

hansung-msft (approved):

ralphgabrinao (approved):

PR #81, opened 2022-08-05
Changes for postCall Config

dailam-msft:
# What
PR for Changes for PostCall surveys -

1.Changes in client\src\models\ConfigModel.ts - to add postCall survey object, optional field in "AppConfigModel"
2. Changes in client\src\utils\FetchConfig.test.ts - to test changes in AppConfigModel and getting config from server.
3.Changes in server\src\models\configModel.ts - to add postCall optional object , in "ServerConfigModel" and "ClientConfigModel"
4.Changes in server\src\defaultConfig.json to map right values
5.Changes in server\src\utils\getConfig.ts - to handle mapping from DefaultConfig object to Client and Server Config models
6.Changes to server\src\utils\getConfig.test.ts - to test changes made .
7. Changes to move getDefaultConfig to separate file for better unit testing -
      - getDefaultConfig.ts
      - getDefaultConfig.test.ts
      - tokenController.test.ts (use above changes)

# Why

- By adding postCall object into defaultConfig.json, we make it known to repo users that this feature is available and how it can be configured to be used.

- Mapping 'type' is included to have 'none' as well as the previously agreed upon 3 other types - msforms, onequestion and thirdparty. 

# How Tested
image

image

image

# Process & policy checklist


server/src/models/configModel.ts
```diff
@@ -1,5 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
+export type PostCallSurveyType = 'none' | 'msforms' | 'onequestion' | 'thirdparty';
```
dailam-msft, line 3: Added 'none' option here since mocking without a postCall object (as per tech doc) threw errors.
image

server/src/utils/getConfig.ts
```diff
@@ -15,33 +15,52 @@ import {
   VV_LOGO_URL_ENV_NAME
 } from '../constants';
 
-import { ServerConfigModel, ClientConfigModel } from '../models/configModel';
+import { ServerConfigModel, ClientConfigModel, PostCallSurveyType } from '../models/configModel';
 import DefaultConfig from '../defaultConfig.json';
 
+export const getDefaultConfig = (): typeof DefaultConfig => {
```
dailam-msft, line 21: Added this and refctored throughout this file to enable mocking and unit testing of defaultConfig.json in getConfig.test.ts

server/src/utils/getConfig.ts
```diff
@@ -15,33 +15,52 @@ import {
...
+        : defaultConfigObj.chatEnabled,
     screenShareEnabled:
       typeof process.env[VV_SCREENSHARE_ENABLED_ENV_NAME] === 'string'
         ? process.env[VV_SCREENSHARE_ENABLED_ENV_NAME]?.toLowerCase() === 'true'
-        : DefaultConfig.screenShareEnabled,
-    companyName: process.env[VV_COMPANY_NAME_ENV_NAME] ?? DefaultConfig.companyName,
-    colorPalette: process.env[VV_COLOR_PALETTE_ENV_NAME] ?? DefaultConfig.colorPalette,
-    waitingTitle: process.env[VV_WAITING_TITLE_ENV_NAME] ?? DefaultConfig.waitingTitle,
-    waitingSubtitle: process.env[VV_WAITING_SUBTITLE_ENV_NAME] ?? DefaultConfig.waitingSubtitle,
-    logoUrl: process.env[VV_LOGO_URL_ENV_NAME] ?? DefaultConfig.logoUrl
```
dailam-msft, line 38: Adding postCall to proccess.env will be done in a separate PR

server/src/utils/getConfig.ts
```diff
@@ -15,33 +15,52 @@ import {
...
+import { ServerConfigModel, ClientConfigModel, PostCallSurveyType } from '../models/configModel';
 import DefaultConfig from '../defaultConfig.json';
 
+export const getDefaultConfig = (): typeof DefaultConfig => {
+  return DefaultConfig;
+};
+
 export const getServerConfig = (): ServerConfigModel => {
-  return {
+  const defaultConfigObj = getDefaultConfig();
```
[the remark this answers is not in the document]
dailam-msft: Hey @hansung-msft I didnt understand the question here.. Was it that I create a model object for defaultConfig.json???

server/src/models/configModel.ts
```diff
@@ -23,4 +36,10 @@ export interface ClientConfigModel {
   waitingTitle: string;
   waitingSubtitle: string;
   logoUrl: string;
+  postCall?: {
+    survey?: {
+      type: PostCallSurveyType;
+      options: MSFormsSurveryOptions | ThirdPartySurveyOptions;
+    };
+  };
```
ralphgabrinao, line 44: This code is repeated. Can we extract this to an interface so we can use the interface in both places?

server/src/controllers/tokenController.test.ts
```diff
@@ -2,9 +2,11 @@
 // Licensed under the MIT license.
 
 import { CommunicationIdentityClient } from '@azure/communication-identity';
-import DefaultConfig from '../defaultConfig.json';
+//import DefaultConfig from '../defaultConfig.json';
```
ralphgabrinao, line 5: We can remove this

server/src/utils/getDefaultConfig.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import DefaultConfig from '../defaultConfig.json';
+
+// for mocking DefaultConfig in unit tests
+export const getDefaultConfig = (): any => DefaultConfig;
```
ralphgabrinao, line 7: In lieu of the `any` type, can we make this a `ServerConfigModel` type instead since it will be serialized to that?

server/src/models/configModel.ts
```diff
@@ -1,5 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
+export type PostCallSurveyType = 'msforms' | 'onequestionpoll' | 'thirdparty';
```
ralphgabrinao, line 3: I'm thinking we shouldn't include `onequestionpoll` yet since it doesn't yet have a corresponding interface and it is not ready to be exposed.

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,69 @@ import {
...
+    config.postCall = {
+      survey: { type: postCallType, options: { surveyUrl: postCallSurveyUrl } }
+    };
+  } else if (defaultConfig.postCall?.survey?.type) {
+    postCallType = <PostCallSurveyType>defaultConfig.postCall?.survey?.type;
+    postCallSurveyUrl = defaultConfig.postCall?.survey?.options?.surveyUrl;
+    config.postCall = {
+      survey: { type: postCallType, options: { surveyUrl: postCallSurveyUrl } }
+    };
+  }
```
ralphgabrinao, line 70: We can simplify/clarify the logic here:
- Set the initial values to the default values
- Try and get each post call environment variable. If found, overwrite the default.

We should also move this logic to a new function for separation of concerns. PoC: https://github.com/Azure-Samples/communication-services-virtual-visits-js/commit/0bbe84a60b65b258419d084bdbcb92499ea039b6#diff-169359eb8c23e178028aa80922c7a185a667bc1732bf59178a66eb2ab78ffdc3R60

In addition, we should write the appropriate unit tests, including a test for handling a value that can't be serialized to a `PostCallSurveyType`

ralphgabrinao (changes requested):

server/src/utils/getDefaultConfig.ts
dailam-msft: This doesnt work because of PostCallType being used, whereas defaultConfig has strings -  
image

server/src/models/configModel.ts
dailam-msft: This change is applied. Thanks!

server/src/models/configModel.ts
dailam-msft: This has been reverrted and 'none' will not b an option in current implementation

server/src/controllers/tokenController.test.ts
dailam-msft: done

server/src/utils/getConfig.ts
dailam-msft: This change has been made

server/src/utils/getConfig.ts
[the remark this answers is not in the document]
hansung-msft: This is outdated! don't worry about it!

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
+  config.postCall = getPostCallConfig(defaultConfig);
+
+  return config;
+};
+
+const isValidPostCallSurveyType = (postcallSurveyType: string): postcallSurveyType is PostCallSurveyType => {
+  return ['msforms', 'thirdparty'].indexOf(postcallSurveyType) !== -1;
+};
+
+const getPostCallConfig = (defaultConfig: any): PostCallConfig | undefined => {
```
hansung-msft, line 54: I am not sure if lint complain about this, but I am not sure it is good idea to leave `any` type for `defaultConfig` parameter.

Since we have `ServerConfigModel` interface and `postCall` property is optional, we can replace `any` to `ServerConfigModel`?

dailam-msft: made this change and pushed . Thanks!

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
+    };
+  }
+
+  try {
+    const postcallSurveyType = process.env[VV_POSTCALL_SURVEY_TYPE_ENV_NAME];
+    const postcallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME];
+
+    if (
+      postcallSurveyType &&
+      (postcallSurveyType === 'msforms' || postcallSurveyType === 'thirdparty') &&
```
ralphgabrinao, line 76: Can we use the same `isValidPostCallSurveyType` function here?

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
+      survey: {
+        type: defaultConfig.postCall?.survey?.type,
+        options: { surveyUrl: defaultConfig.postCall?.survey?.options?.surveyUrl }
+      }
+    };
+  }
+
+  try {
+    const postcallSurveyType = process.env[VV_POSTCALL_SURVEY_TYPE_ENV_NAME];
+    const postcallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME];
```
ralphgabrinao, line 72: I think this try/catch is a great starting point. The code from lines 57-68 very closely resembles the code in this try catch. Can we combine them like so?

```
const postCallSurveyType = process.env[VV_POSTCALL_SURVEY_TYPE_ENV_NAME] ?? defaultConfig.postCall?.survey?.type;
const postCallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME] ?? defaultConfig.postCall?.survey?.options?.surveyUrl;

...
rest of validation logic
...
```

This way we just have one single code block that does the validation instead of two.

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
+      postcallSurveyType &&
+      (postcallSurveyType === 'msforms' || postcallSurveyType === 'thirdparty') &&
+      postcallSurveyUrl
+    ) {
+      postcallConfig = {
+        survey: { type: postcallSurveyType, options: { surveyUrl: postcallSurveyUrl } }
+      };
+    }
+  } catch (e) {
+    console.error('Unable to parse post call values from environment variables');
```
ralphgabrinao, line 84: I think throwing here makes sense. The app cannot start because the postCall config is invalid.

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
-  };
+        : defaultConfig.screenShareEnabled,
+    companyName: process.env[VV_COMPANY_NAME_ENV_NAME] ?? defaultConfig.companyName,
+    colorPalette: process.env[VV_COLOR_PALETTE_ENV_NAME] ?? defaultConfig.colorPalette,
+    waitingTitle: process.env[VV_WAITING_TITLE_ENV_NAME] ?? defaultConfig.waitingTitle,
+    waitingSubtitle: process.env[VV_WAITING_SUBTITLE_ENV_NAME] ?? defaultConfig.waitingSubtitle,
+    logoUrl: process.env[VV_LOGO_URL_ENV_NAME] ?? defaultConfig.logoUrl
+  } as ServerConfigModel;
+
+  config.postCall = getPostCallConfig(defaultConfig);
```
ralphgabrinao, line 45: We can put this in the object directly instead of a separate assignment.

server/src/utils/getConfig.ts
```diff
@@ -51,5 +99,17 @@ export const getClientConfig = (serverConfig: ServerConfigModel): ClientConfigMo
     waitingTitle: serverConfig.waitingTitle,
     waitingSubtitle: serverConfig.waitingSubtitle,
     logoUrl: serverConfig.logoUrl
-  };
+  } as ClientConfigModel;
+
+  if (serverConfig.postCall?.survey?.type) {
```
ralphgabrinao, line 104: At this point, the validations have already happened in `getServerConfig`. IMO we should just assign it directly:

```
postCall: serverConfig.postCall
```

server/src/utils/getDefaultConfig.ts
ralphgabrinao: I am certain it is possible, based on this PoC: https://github.com/Azure-Samples/communication-services-virtual-visits-js/commit/0bbe84a60b65b258419d084bdbcb92499ea039b6#diff-434fe846a2edd7e4f0218b939c7d4f20f6b7724eda2d61d7dcc27e59cf3c9314

Is it because we are assigning blank values in the postCall property in defaultConfig and it is inferring the type? Can we assign null or undefined instead?

Ideally we would like to avoid `any` especially in this case since the config is the basis of the entire app and we'd want to be alerted of any issues at compile-time.

server/src/utils/getConfig.ts
```diff
@@ -12,36 +12,84 @@ import {
...
+      }
+    };
+  }
+
+  try {
+    const postcallSurveyType = process.env[VV_POSTCALL_SURVEY_TYPE_ENV_NAME];
+    const postcallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME];
+
+    if (
+      postcallSurveyType &&
```
ralphgabrinao, line 75: Can we move this null check to the `isValidPostCallSurveyType` function so that all validations are consolidated?

server/src/utils/getDefaultConfig.ts
dailam-msft: The one in prototype worked since at that point we thought entire postCall object would be null. Assigning null and undefined to type in defaultConfig.json doesnt work either - 
image
image
I understan trying to avoid any, but this is somethign that we can address/fix (will probably fix itself ) if and when onequestionpoll is implemented and becomes the default

server/src/utils/getConfig.ts
dailam-msft: TRied that , doesnt work. ALso cant give it as string | undefined coz indexof function will not work if input parameter is given as undefined - 
image

server/src/utils/getConfig.ts
dailam-msft: But postcall is optional.. so if it is undefined/invalid throwing makes sense yes, but the app would still start and the rest would work BAU i think.

ralphgabrinao: Great point

server/src/defaultConfig.json
```diff
@@ -7,5 +7,13 @@
   "colorPalette": "#0078d4",
   "waitingTitle": "Thank you for choosing Lamna Healthcare",
   "waitingSubtitle": "Your clinician is joining the meeting",
-  "logoUrl": ""
+  "logoUrl": "",
+  "postCall": {
+    "survey": {
+      "type": "",
```
ralphgabrinao, line 13: Would this cause the validation to fail? Empty string is not a valid type

server/src/utils/getConfig.ts
hansung-msft: I am wondering if we just assign it directly and `serverConfig.postCall` is `undefined`, the config will be formulate as below:

image

Is this what we intended to?

server/src/defaultConfig.json
```diff
@@ -7,5 +7,13 @@
...
+      "options": {
+        "surveyUrl": ""
```
ralphgabrinao, line 15: Setting `surveyUrl` here seems odd. This options object with surveyUrl implements a specific interface (`MSFormsSurveyOptions` and `ThirdPartySurveyOptions`), and yet we are not setting the type as MS Forms or Third Party. Is it more appropriate to set `options` as null or undefined? Or set the entire `postCall` as null or undefined?

server/src/utils/getConfig.ts
ralphgabrinao: I think so. The client will render the default post call UI in that case.

hansung-msft (approved):
Once Ralph's comments resolves, it looks good to me!

server/src/utils/getConfig.ts
hansung-msft: I thought since `postCall` is optional that `postCall` property shouldn't have to be in `config` if it the value doesn't exist.

hansung-msft: After we discussed offline, at this stage, it will be okay with `postCall: undefined`.

ralphgabrinao (approved):

hansung-msft (approved):

PR #86, opened 2022-08-26
survey fhl integation into current sample

dailam-msft:
# What
First integration of FHL postcall into sample code. 
The following is being worked upon - 
 - Pass in config (remove separate fetchConfig api call within survey component) and bring in checks wether SurveyForm or Noticepage is to be shown

# Why
Brainstorming on ways to implement "re-join call"


dailam-msft:
CLosing this DRAFT PR

PR #87, opened 2022-09-01
Adding Survey component to be displayed after end call

dailam-msft:
# What
Adding Survey component and rejoin call link,  to be displayed after the call ends 

# Why
To implement PostCallSurvey feature

# How Tested
tested on local . SS below - 
1) When defaultConfig.json has postcallconfig empty (option "none") is selected - 
image

2) When msforms is selected option
image

3)On click of "rejoin call" link, user is taken back to lobby - 
image

# Process & policy checklist


client/src/models/ConfigModel.ts
```diff
@@ -3,13 +3,20 @@
...
   surveyUrl: string;
 }
 interface ThirdPartySurveyOptions {
   surveyUrl: string;
 }
+
+export interface PostCallConfig {
+  survey?: {
+    type: PostCallSurveyType;
+    options: MSFormsSurveryOptions | ThirdPartySurveyOptions;
```
tejaswip-msft, line 17: typo, there is an extra "r" in the survey
should be MsFormsSurveyOptions

client/src/components/MeetingExperience.tsx
```diff
@@ -75,12 +89,26 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
...
     const locale = COMPOSITE_LOCALE_EN_US;
     const formFactorValue = new MobileDetect(window.navigator.userAgent).mobile() ? 'mobile' : 'desktop';
 
+    if (renderPostCall) {
+      return (
+        <Survey
+          postCall={postCall}
+          onRejoinCall={() => {
+            callWithChatAdapter.joinCall();
+            callWithChatAdapter.onStateChange((state: CallWithChatAdapterState) => {
```
mbakalov, line 103: Feels like the order of methods should be changed? First, we subscribe to events that happen when we `joinCall` and then we initiate the actual `joinCall`.

client/src/components/MeetingExperience.tsx
```diff
@@ -75,12 +89,26 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
...
+              if (state.page === 'configuration') state.page = 'lobby';
```
mbakalov, line 104: We should probably add a comment about why this is necessary.

client/src/components/MeetingExperience.tsx
```diff
@@ -64,6 +73,11 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
           endpointUrl,
           chatEnabled
         );
+        adapter.on('callEnded', async () => {
```
mbakalov, line 76: Doesn't matter much but probably doesn't need to be `async` since we're not awaiting anything?

client/src/components/Survey.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }
+};
+
+describe('Survey', () => {
+  it('should render', async () => {
+    const survey = await mount<SurveyProps>(<Survey postCall={mockPostCall} onRejoinCall={jest.fn()} />);
+    expect(survey.find('iframe').length).toBe(1);
+    expect(survey.props().postCall?.survey?.type).toBe(mockPostCall.survey?.type);
+    expect(survey.props().postCall?.survey?.options.surveyUrl).toBe(mockPostCall.survey?.options.surveyUrl);
```
mbakalov, line 25: Not sure how useful it is: we are setting the prop and then checking that the prop was set correctly. Are we testing any code that we wrote as part of this? It doesn't seem so.

A more useful check would be to verify that iframe's url is set to the value we expect.

client/src/components/MeetingExperience.tsx
mbakalov: More importantly, if we don't have any postCall, then maybe we shouldn't subscribe to `callEnded` at all? Instead of always subscribing and then checking if we actually need the survey in the handler.

client/src/components/Survey.tsx
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license.
+
+import React from 'react';
+import { Link, Stack } from '@fluentui/react';
+import { PostCallConfig } from '../models/ConfigModel';
+import { rejoinLinkStyle, surveyStyle } from '../styles/Survey.styles';
+
+export interface SurveyProps {
+  onRejoinCall: () => void;
+  postCall?: PostCallConfig;
```
mbakalov, line 11: Since we only want to render the survey when postCall actually exists, then I think it is better we make this non-nullable.

dailam-msft: Since we are flowing down the config from AppConfigModel->Visit.tsx->MeetingExperience.tsx->Survey.tsx , I declared it as an optional field. Making this mandatory here throws compile errors in places where it is being assigned . 

AppConfigModel has postCall as optional and so in other components using it too, it's marked optional. 

I am making the change to have the if condition checking the postCallType before the subscription to onCallEnded event on adapter and am in the process of testing it. That will make sure that Survey is rendered only in the postcall scenarios. What do you think?

client/src/components/MeetingExperience.tsx
dailam-msft: Tested this change and it works both ways. But yes, reads better to subscribe first and trigger after.

client/src/components/MeetingExperience.tsx
dailam-msft: Thanks, have made this change.

client/src/components/MeetingExperience.tsx
dailam-msft: have made this change.

client/src/components/MeetingExperience.tsx
dailam-msft: have made this change.

client/src/components/Survey.tsx
mbakalov: It is annoying to have all the nullability checks (`props.postCall?.survey?.options.surveyUrl`) in the `Survey` component. Instead could `Survey`'s props interface demand that `postCall` be non-nullable? Then whoever calls the `Survey` needs to provide proof that it has non-null `postCall`:

```typescript
if (renderPostCall && postCall) {
  return (
    <Survey postCall={postCall} ... />
  );
}
```

client/src/models/ConfigModel.ts
```diff
@@ -3,13 +3,20 @@
...
+export type PostCallSurveyType = 'msforms' | 'thirdparty';
+interface MSFormsSurveyOptions {
   surveyUrl: string;
 }
 interface ThirdPartySurveyOptions {
   surveyUrl: string;
 }
+
+export interface PostCallConfig {
+  survey?: {
```
mbakalov, line 15: Should this be optional? `postCall` itself is optional, and we don't have any post-calls without survey, right?

client/src/components/MeetingExperience.tsx
```diff
@@ -46,11 +54,12 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
     userId,
     waitingSubtitle,
     waitingTitle,
+    postCall,
     onDisplayError
   } = props;
 
   const [callWithChatAdapter, setCallWithChatAdapter] = useState<CallWithChatAdapter | undefined>(undefined);
-
+  const [renderPostCall, setRenderPostCall] = useState(false);
```
hansung-msft, line 62: Can we put type in for `useState`?

`const [renderPostCall, setRenderPostCall] = useState<boolean>(false);`

client/src/components/Survey.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+const SURVEY = 'SurveyComponent';
+
+export const Survey: React.FunctionComponent<SurveyProps> = (props: SurveyProps) => {
+  return (
+    <Stack styles={surveyStyle}>
+      <iframe
+        title={SURVEY}
+        style={{ height: '100%', width: '100%', border: 0 }}
```
hansung-msft, line 20: I think it would better to move this to separate variable for consistency.

client/src/models/ConfigModel.ts
dailam-msft: This was brought up during the tech doc discussion as well as later and the thought process explained was that we may have "other post call activities" and not just survey.

client/src/models/ConfigModel.ts
dailam-msft: This is done

client/src/components/Survey.tsx
dailam-msft: Thanks! Have made this change.

client/src/components/MeetingExperience.tsx
dailam-msft: Thanks! Have made this change..

client/src/components/Survey.tsx
dailam-msft: Thanks, have made this change

client/src/components/Survey.test.tsx
dailam-msft: Have made this change, the iframe's url will now be tested to match the one passed in on props of Survey

client/src/components/MeetingExperience.tsx
dailam-msft: From UX review it is seen that when we end call on "Joining Call" screen (lobby hasnt fully loaded, the survey page gets skipped. This behaviour is avoided when joinCall is called first and subscribing to onStateChanged happens right after. So I think it would be better to revert.

dailam-msft: Ughh.. actually that doesnt fix it. Will have to do some more digging to see what a potential fix might be

client/src/components/MeetingExperience.tsx
```diff
@@ -31,9 +34,14 @@ export interface MeetingExperienceProps {
   waitingSubtitle: string;
   logoUrl: string;
   chatEnabled: boolean;
+  postCall: PostCallConfig | undefined;
   onDisplayError(error: any): void;
 }
 
+const isValidPostCallSurveyType = (postcallSurveyType: string): postcallSurveyType is PostCallSurveyType => {
```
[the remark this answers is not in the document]
dailam-msft: After some reading and discussion, the conslusion is that though typescript can enforce the type during compile-time, at runtime this doesnt happen and so the check is needed

mbakalov (approved):

hansung-msft (approved):

mbakalov (approved):

PR #91, opened 2022-09-16
Config changes for post-call survey V2

tsohoni-msft:
# What

- Adding a new survey type and related options in existing post-call config.
- Making changes to fetch the config for the new survey type.
- Adding variables to ARM templates.
- Fixing tests.

# Why
Scaffolding for future V2 work by implementing the config.

# How Tested
Ran all unit tests, tested ARM.

# Process & policy checklist


server/src/defaultConfig.json
```diff
@@ -7,5 +7,11 @@
...
   "waitingTitle": "Thank you for choosing Lamna Healthcare",
   "waitingSubtitle": "Your clinician is joining the meeting",
-  "logoUrl": ""
+  "logoUrl": "",
+  "postCallSurveyType": "",
+  "postCallSurveyOptionsSurveyUrl": "",
+  "postCallSurveyOneQuestionPollTitle": "",
+  "postCallSurveyOneQuestionPollPrompt": "",
+  "postCallSurveyOneQuestionPollPromptType": "",
+  "postCallSurveyOneQuestionPollSaveButtonText": ""
```
ralphgabrinao, line 16: We can use the nested syntax here and provide the appropriate default values:

```
"postCall": {
    "survey": {
        "type": "onequestionpoll",
        "options": {
            "pollTitle": "Tell us how we did",
            "pollPrompt": "How satisfied are you with this virtual appointment’s audio and video quality?"
            ....
        }
    }
}
```

ralphgabrinao (changes requested):

client/src/components/Survey.tsx
```diff
@@ -13,14 +13,20 @@ export interface SurveyProps {
...
+  }
   return (
     <Stack styles={surveyStyle}>
-      <iframe
-        title={SURVEY}
-        style={surveyIframeStyle}
-        src={props.postCall.survey?.options.surveyUrl}
-        scrolling="yes"
-      ></iframe>
+      <iframe title={SURVEY} style={surveyIframeStyle} src={postcallSurveyUrl} scrolling="yes"></iframe>
```
ralphgabrinao, line 29: I think we first have to build the UI for one question poll before we make it the default option in the config, if I am not mistaken. Otherwise, when the client gets the config and sees that post call is set to one question poll, it doesn't have yet the UI to handle that.

server/src/defaultConfig.json
tsohoni-msft: As per discussion with Ralph over chat, reverted this file along with ARM changes for now since we want to include those in a separate PR once the other UI work for V2 is done and it is ready for releasing.

client/src/components/Survey.tsx
tsohoni-msft: Yep, I agree, so we will not make one question poll the default option for now.

server/src/utils/getConfig.ts
```diff
@@ -56,13 +69,43 @@ const getPostCallConfig = (defaultConfig: ServerConfigModel): PostCallConfig | u
...
-    const postcallSurveyUrl =
-      process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME] ?? defaultConfig.postCall?.survey.options?.surveyUrl;
+    let postcallSurveyUrl = '';
+    if (postcallSurveyType === 'msforms') {
+      const options: MSFormsSurveyOptions = defaultConfig.postCall?.survey?.options as MSFormsSurveyOptions;
+      postcallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME] ?? options.surveyUrl;
+    } else if (postcallSurveyType === 'custom') {
+      const options: CustomSurveyOptions = defaultConfig.postCall?.survey?.options as CustomSurveyOptions;
+      postcallSurveyUrl = process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME] ?? options.surveyUrl;
+    } else {
```
ralphgabrinao, line 79: Making this `else if (postCallSurveyType === 'onequestionpoll')` is better I think. Then inside that block is where we create the 1QP Config object.

Then, at the end is the `else` block where we should return `undefined` which is basically setting postcall to "None". At that point, it means that postCallSurveyType did not match any of `msforms`, `custom` or `onequestionpoll`

ralphgabrinao: If it helps make the code more concise/readable, I suggest creating functions that return the right options object based on survey type:

`const createMSFormsOptions = (): MSFormsSurveyOptions => { ... }`
`const createCustomFormOptions = (): CustomSurveyOptions => { ... }`
`const createOneQuestionPollOptions = (): OneQuestionPollOptions => { ... }`

```
if (postcallSurveyType === 'msforms') {
    const options: MSFormsSurveyOptions = createMSFormsOptions();
    return {
        survey: {
            type: 'msforms',
            options: options
        }
    }
}
```

client/src/models/ConfigModel.ts
```diff
@@ -3,18 +3,27 @@
...
   surveyUrl: string;
 }
 
+export type OneQuestionPollPromptType = 'likeorDislike' | 'rating' | 'text';
+
+export interface OneQuestionPollOptions {
+  title?: string;
+  prompt: string;
+  promptType: OneQuestionPollPromptType;
+  saveButtonText: string;
```
tejaswip-msft, line 20: we would need answer place holder also included in this right
This comes only if the option is "text input"

client/src/models/ConfigModel.ts
```diff
@@ -3,18 +3,27 @@
...
+export type PostCallSurveyType = 'msforms' | 'custom' | 'onequestionpoll';
+export interface MSFormsSurveyOptions {
   surveyUrl: string;
 }
-interface CustomSurveyOptions {
+export interface CustomSurveyOptions {
   surveyUrl: string;
 }
 
+export type OneQuestionPollPromptType = 'likeorDislike' | 'rating' | 'text';
```
tejaswip-msft, line 14: lets name this OneQuestionPollType instead

client/src/models/ConfigModel.ts
```diff
@@ -3,18 +3,27 @@
...
+
+export interface OneQuestionPollOptions {
+  title?: string;
+  prompt: string;
+  promptType: OneQuestionPollPromptType;
```
tejaswip-msft, line 19: naming here as well, pollType: OneQuestionPollType

server/src/utils/getConfig.ts
tsohoni-msft: Agreed, made changes accordingly.

client/src/models/ConfigModel.ts
tsohoni-msft: Not sure why we would need a placeholder here? The placeholder is only needed in the defaultConfig right?

client/src/models/ConfigModel.ts
tsohoni-msft: Agreed, done!

client/src/models/ConfigModel.ts
tsohoni-msft: Agreed, done!

deploy/azuredeploy.json
```diff
@@ -115,7 +119,11 @@
             "VV_WAITING_SUBTITLE": "[variables('waitingSubtitle')]",
             "VV_LOGO_URL": "[variables('logoUrl')]",
             "VV_POSTCALL_SURVEY_TYPE": "[variables('postCallSurveyType')]",
-            "VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL": "[variables('postCallSurveyOptionsSurveyUrl')]"
+            "VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL": "[variables('postCallSurveyOptionsSurveyUrl')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_TITLE_ENV_NAME": "[variables('postCallSurveyOneQuestionPollTitle')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_PROMPT_ENV_NAME": "[variables('postCallSurveyOneQuestionPollPrompt')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_TYPE_ENV_NAME": "[variables('postCallSurveyOneQuestionPollType')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME": "[variables('postCallSurveyOneQuestionPollSaveButtonText')]"
```
ralphgabrinao, line 126: Looks like the env vars should not have `_ENV_NAME` at the end of them

deploy/azuredeployexistingresource.json
```diff
@@ -77,7 +81,11 @@
             "VV_WAITING_SUBTITLE": "[variables('waitingSubtitle')]",
             "VV_LOGO_URL": "[variables('logoUrl')]",
             "VV_POSTCALL_SURVEY_TYPE": "[variables('postCallSurveyType')]",
-            "VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL": "[variables('postCallSurveyOptionsSurveyUrl')]"
+            "VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL": "[variables('postCallSurveyOptionsSurveyUrl')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_TITLE_ENV_NAME": "[variables('postCallSurveyOneQuestionPollTitle')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_PROMPT_ENV_NAME": "[variables('postCallSurveyOneQuestionPollPrompt')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_TYPE_ENV_NAME": "[variables('postCallSurveyOneQuestionPollType')]",
+            "VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME": "[variables('postCallSurveyOneQuestionPollSaveButtonText')]"
```
ralphgabrinao, line 88: Same comment re: `_ENV_NAME`

server/src/utils/getConfig.ts
```diff
@@ -14,12 +14,54 @@ import {
...
+  PostCallSurveyType,
+  PostCallConfig,
+  MSFormsSurveyOptions,
+  OneQuestionPollOptions,
+  OneQuestionPollType,
+  CustomSurveyOptions
+} from '../models/configModel';
 import { getDefaultConfig } from './getDefaultConfig';
 
+const getMSFormsOptions = (defaultConfig: any): MSFormsSurveyOptions => {
```
ralphgabrinao, line 36: Can we hard type this to `ServerConfigModel` instead of `any` ?

server/src/utils/getConfig.ts
```diff
@@ -56,13 +98,25 @@ const getPostCallConfig = (defaultConfig: ServerConfigModel): PostCallConfig | u
...
   try {
     const postcallSurveyType = process.env[VV_POSTCALL_SURVEY_TYPE_ENV_NAME] ?? defaultConfig.postCall?.survey.type;
-    const postcallSurveyUrl =
-      process.env[VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL_ENV_NAME] ?? defaultConfig.postCall?.survey.options?.surveyUrl;
-
-    if (postcallSurveyType && isValidPostCallSurveyType(postcallSurveyType) && postcallSurveyUrl) {
-      postcallConfig = {
-        survey: { type: postcallSurveyType, options: { surveyUrl: postcallSurveyUrl } }
-      };
+    if (postcallSurveyType && isValidPostCallSurveyType(postcallSurveyType)) {
```
ralphgabrinao, line 101: Minor nit: We can reduce the nesting by short circuiting here:

```
if (!postCallSurveyType || !isValidPostCallSurveyType(postCallSurveyType)) {
    return undefined;
}
```

server/src/utils/getConfig.test.ts
```diff
@@ -138,13 +141,16 @@ describe('config', () => {
       .spyOn(getDefaultConfig, 'getDefaultConfig')
       .mockImplementation((): any => mockDefaultConfig);
 
-    const config = getConfig.getServerConfig();
+    const serverConfig = getConfig.getServerConfig();
+    serverConfig.communicationServicesConnectionString = 'endpoint=test_endpoint_value;accesskey=secret';
+    const clientConfig = getConfig.getClientConfig(serverConfig);
```
ralphgabrinao, line 146: Hmm not sure why we needed to add this call here. The test is for `getServerConfig`

server/src/utils/getConfig.test.ts
```diff
@@ -138,13 +141,16 @@ describe('config', () => {
...
-    expect(config.postCall?.survey.options).toBeDefined();
-    expect(config.postCall?.survey.options?.surveyUrl).toBe(mockDefaultConfig.postCall.survey.options.surveyUrl);
+    expect(clientConfig.postCall).toBeDefined();
+    expect(clientConfig.postCall?.survey).toBeDefined();
+    expect(clientConfig.postCall?.survey.type).toBe(mockDefaultConfig.postCall.survey.type);
+    expect(clientConfig.postCall?.survey.options).toBeDefined();
+    expect(options.surveyUrl).toBe(mockDefaultConfig.postCall.survey.options.surveyUrl);
   });
 
   test('getServerConfig returns empty postcall object when postCallSurveyType is invalid', () => {
```
ralphgabrinao, line 156: Nit: I think we should change the test name to be more accurate. `getServerConfig returns undefined when postCallSurveyType is invalid`

server/src/utils/getConfig.ts
```diff
@@ -56,13 +98,25 @@ const getPostCallConfig = (defaultConfig: ServerConfigModel): PostCallConfig | u
...
+        const configOptions: MSFormsSurveyOptions = getMSFormsOptions(defaultConfig);
+        postcallConfig = {
+          survey: { type: postcallSurveyType, options: configOptions }
+        };
+      } else if (postcallSurveyType === 'custom') {
+        const configOptions: CustomSurveyOptions = getCustomSurveyOptions(defaultConfig);
+        postcallConfig = {
+          survey: { type: postcallSurveyType, options: configOptions }
+        };
+      } else if (postcallSurveyType === 'onequestionpoll') {
```
ralphgabrinao, line 112: I think we are missing some tests cases.

From what I can tell we need at least two test cases per postCallSurveyType:
- one that uses env vars
- one that uses defaultConfig

In other words, we need to test the logic inside of `getMSFormsOptions`, `getCustomSurveyOptions`, and `getOneQuestionPollOptions`.

client/src/models/ConfigModel.ts
tejaswip-msft: answer placeholder is also a user input, but is shown only if the one question poll option is "Text input"
here is the figma: https://www.figma.com/file/zinLSNSrpQGoV6Hmbw1t1C/Virtual-visit-add-ons?node-id=424%3A72988

deploy/azuredeploy.json
tsohoni-msft: Fixed this.

server/src/utils/getConfig.ts
tsohoni-msft: Done.

hansung-msft:
Are we include `VV_COSMOS_DB_ENDPOINT` and `VV_COSMOS_DB_NAME` configs in this PR?

server/src/models/configModel.ts
```diff
@@ -1,17 +1,26 @@
...
+export type PostCallSurveyType = 'msforms' | 'custom' | 'onequestionpoll';
+export interface MSFormsSurveyOptions {
   surveyUrl: string;
 }
-interface CustomSurveyOptions {
+export interface CustomSurveyOptions {
   surveyUrl: string;
 }
+
+export type OneQuestionPollType = 'likeorDislike' | 'rating' | 'text';
```
ralphgabrinao, line 12: Should we capitalize the 'o' in `likeOrDislike`?

hansung-msft:
> Are we include `VV_COSMOS_DB_ENDPOINT` and `VV_COSMOS_DB_NAME` configs in this PR?

Ah forget about this comment. Tejaswi told me there is a user story for that. Sorry for the confusion.

client/src/models/ConfigModel.ts
tsohoni-msft: Added this to the model.

deploy/azuredeployexistingresource.json
tsohoni-msft: Fixed this.

server/src/utils/getConfig.test.ts
tsohoni-msft: That was my bad, forgot to revert from something I was trying. Fixed this now.

server/src/utils/getConfig.test.ts
tsohoni-msft: Fixed this.

server/src/models/configModel.ts
tsohoni-msft: Fixed this.

server/src/utils/getConfig.ts
```diff
@@ -14,12 +14,54 @@ import {
...
+    (process.env[VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_TYPE_ENV_NAME] as OneQuestionPollType) ??
+    (options.pollType as OneQuestionPollType);
+  const surveyPrompt = process.env[VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_PROMPT_ENV_NAME] ?? options.prompt;
+  const surveySaveButtonText =
+    process.env[VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME] ?? options.saveButtonText;
+  return {
+    title: surveyTitle,
+    prompt: surveyPrompt,
+    pollType: surveyPromptType,
+    saveButtonText: surveySaveButtonText
```
ralphgabrinao, line 61: Missing `answerPlaceholder` here

tsohoni-msft: My bad, missed adding stuff related to the answerPlaceholder. Fixed that now.

ralphgabrinao (approved):

tejaswip-msft (approved):

PR #92, opened 2022-09-19
Changes according to recomendation by Web UI team

dailam-msft:
PR to change Survey and CallWithChatComponent rendering according to WebUI team's recommendations

Suggestion was made that we should not be setting state.page to 'lobby' directly.

The page was changing to 'configuration' because CallWithChatComposite was unmounted. 

The change now is that CallWithChatComposite will have display set to 'none' if renderPostCall is true and Survey component will be shown instead.

Tested locally - 
image
image


client/src/components/MeetingExperience.test.tsx
```diff
@@ -187,10 +187,13 @@ describe('MeetingExperience', () => {
...
+    expect(setRenderPostCallMock).toBeCalled();
     const callWithChatComposites = meetingExperience.find(CallWithChatComposite);
-    expect(callWithChatComposites.length).toBe(0);
-    expect(survey.length).toBe(1);
-    expect(callWithChatComposites.length).toBe(0);
+    expect(callWithChatComposites.length).toBe(1);
+    const parentDiv = callWithChatComposites.parent();
+    expect(parentDiv.props().style.display).toBe('none');
+
+    const iframe = meetingExperience.find('iframe');
```
ralphgabrinao, line 196: Can we find the Survey component directly here?
```
const surveyComponent = meetingExperience.find(Survey);
expect(surveyComponent.length).toBe(1);
```

That way the iframe is abstracted from this test.

ralphgabrinao: i see we do the inverse of this in the previous test above line 160, so making this consistent makes sense

client/src/components/MeetingExperience.tsx
```diff
@@ -89,49 +88,47 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
...
-          }
-        }}
-        icons={{
-          LobbyScreenWaitingToBeAdmitted: logo,
-          LobbyScreenConnectingToCall: logo
-        }}
-        formFactor={formFactorValue}
-      />
+      <>
+        {renderPostCall && postCall && (
```
ralphgabrinao, line 93: We should expand the test to cover the case where `postCall` is falsy and assert that `Survey` is not rendered.

client/src/components/MeetingExperience.tsx
```diff
@@ -89,49 +88,47 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
...
+          <Survey
+            data-testid="Survey"
+            postCall={postCall}
+            onRejoinCall={async () => {
+              await callWithChatAdapter.joinCall();
+              setRenderPostCall(false);
+            }}
+          />
+        )}
+        <div style={{ display: renderPostCall ? 'none' : 'flex', flexGrow: 1, position: 'relative', height: '100%' }}>
```
ralphgabrinao, line 103: We should match the conditions here as line 93 since the visibility of Survey and CallWithChatComposite are mutually exclusive. `renderPostCall && postCall`.

client/src/components/MeetingExperience.tsx
dailam-msft: 'should render CallWithChatComposite when renderPostCall is false' handles this.. renderPostCall being false by default, will get set to true only if postCall has some value for "type" . The reason this "&& postCall' check was added was since postCall was optional, but was being passed as prop to Survey component.

client/src/components/MeetingExperience.test.tsx
dailam-msft: made this change

client/src/components/MeetingExperience.tsx
dailam-msft: renderPostCall is false by default, gets set to true only if postCall.survey.type has a value. The reason the "&&postCall" check is added was because postCall is an optinal prop. this is not needed at this point and is not being used for the render of CallWithChatComposite

client/src/components/MeetingExperience.tsx
ralphgabrinao: I am not sure about that. We don't have a test if postCall is undefined:
```
      <MeetingExperience
        userId={{ communicationUserId: 'test' }}
        token={'token'}
        displayName={'name'}
        endpointUrl={'endpoint'}
        locator={{ meetingLink: 'meeting link' }}
        fluentTheme={undefined}
        waitingTitle={waitingTitle}
        waitingSubtitle={waitingSubtitle}
        logoUrl={logoUrl}
        chatEnabled={true}
        postCall={undefined} <----
        onDisplayError={jest.fn()}
      />
```

ralphgabrinao: Yes I see what you mean that it is protected by this condition `if (postCall?.survey.type)`, but we still need to add a unit test for it.

dailam-msft: done

client/src/components/MeetingExperience.test.tsx
```diff
@@ -187,10 +187,48 @@ describe('MeetingExperience', () => {
...
+    expect(callWithChatComposites.length).toBe(1);
+    const parentDiv = callWithChatComposites.parent();
+    expect(parentDiv.props().style.display).toBe('none');
+
+    const survey = meetingExperience.find(Survey);
     expect(survey.length).toBe(1);
-    expect(callWithChatComposites.length).toBe(0);
+  });
+
+  it('should not render Survey component', async () => {
```
ralphgabrinao, line 200: nit: Under what condition does this apply?

Same with the test case before, I think we should add more context to the test names:
- should render Survey component if postCall is defined
- should not render Survey component if postCall is undefined

client/src/components/MeetingExperience.test.tsx
```diff
@@ -187,10 +187,48 @@ describe('MeetingExperience', () => {
...
+
+    meetingExperience.update();
+
+    expect(setRenderPostCallMock).toBeCalled();
+    const survey = meetingExperience.find(Survey);
+    expect(survey.length).toBe(0);
+    const callWithChatComposites = meetingExperience.find(CallWithChatComposite);
+    expect(callWithChatComposites.length).toBe(1);
+    const parentDiv = callWithChatComposites.parent();
+    expect(parentDiv.props().style.display).toBe('flex');
```
ralphgabrinao, line 232: Nice

client/src/components/MeetingExperience.tsx
```diff
@@ -89,49 +88,54 @@ export const MeetingExperience = (props: MeetingExperienceProps): JSX.Element =>
...
+            data-testid="Survey"
+            postCall={postCall}
+            onRejoinCall={async () => {
+              await callWithChatAdapter.joinCall();
+              setRenderPostCall(false);
+            }}
+          />
+        )}
+        <div
+          style={{
```
hansung-msft, line 104: I think we can separate this style properties to `MeetingExperience.styles.ts` file.

dailam-msft: Fixed this

client/src/components/MeetingExperience.test.tsx
dailam-msft: Fixed this

ralphgabrinao (approved):

hansung-msft (approved):

PR #93, opened 2022-09-21
Add readme for post call config

ralphgabrinao:
# What
New md file for post-call survey configuration

# Why
A dedicated section for post-call survey will make it easier to know the different configurations for post-call. Soon the md file will be expanded to include one question poll.

# How Tested

# Process & policy checklist

### DIFF
--- a/docs/post-call-survey.md
+++ b/docs/post-call-survey.md
@@ -6,6 +6,15 @@ You can configure post-call surveys to collect valuable feedback like quality of
 
 Render your [Microsoft Form](https://www.microsoft.com/en-ca/microsoft-365/online-surveys-polls-quizzes) after a call ends:
 
+### Configure using environment variables
+
+```
+VV_POSTCALL_SURVEY_TYPE = "msforms"
+VV_POSTCALL_SURVEY_OPTIONS_SURVEYURL = "<your_MSForms_url>"
+```
+
+### Configure using defaultConfig.json
+
 ```
 {
     ...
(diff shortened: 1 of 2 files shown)


docs/post-call-survey.md
```diff
@@ -0,0 +1,69 @@
...
+
+The [Azure Communication Services UI Library](https://azure.github.io/communication-ui-library/) provides a default post-call screen that is rendered after a call ends, no configuration required.
+
+### Configure using environment variables
+
+Remove `VV_POSTCALL_SURVEY_TYPE` from your environment variables.
+
+### Configure using defaultConfig.json
+
+Remove `postCall` from your config.
```
tejaswip-msft: there wont be any 'postCall' in the config in this case right, so can we just say that leave the default config as is

ralphgabrinao: That's true, though I think by saying 'remove' or 'delete' explicitly tells the user that postCall should not be set. Also would apply once one question poll is added.

tejaswip-msft: Yup makes more sense once one question poll is enabled

docs/post-call-survey.md
```diff
@@ -0,0 +1,63 @@
...
+                "surveyUrl": "<your_survey_url>"
+            }
+        }
+    }
+}
+```
+
+## Use the Default Post-Call Screen
+
+The [Azure Communication Services UI Library](https://azure.github.io/communication-ui-library/) provides a default post-call screen that is rendered after a call ends, no configuration required.
```
tejaswip-msft, line 61: is it okay that these links open in the same tab?

ralphgabrinao: I think so. That's also the case for the links in the main README. Apparently there is no way to open link in new tab in GitHub markdown: https://stackoverflow.com/a/42156438

tejaswip-msft (approved):

ralphgabrinao:
@hansung-msft -- 1QP Will be added later when the feature is completed

hansung-msft (approved):

PR #94, opened 2022-09-22
Hansung/one question poll api test

hansung-msft:
# This is Draft PR for testing the API.
# What

- This PR includes implementation of create survey result API.
- This **Test** branch includes the frontend that you can simply send the test API to the backend. You can modify the payload in the API with that.

# How Tested

```
specs:
node -v: 16.15.0
npm -v: 8.5.5
```
1. [Download Azure Cosmos DB Emulator](https://learn.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21) on the local.
2. Run `npm i` under `/server` directory.
3. Set `communicationServicesConnectionString` in `defaultConfig.json` file.
4. Add following environment variables:  
    ```
      VV_COSMOS_DB_CONNECTION_STRING: <your_db_connection_string>
      VV_COSMOS_DB_ENDPOINT: <your_db_endpoint>
      VV_COSMOS_DB_NAME: <your_db_name>
    ```
    You can find `VV_COSMOS_DB_CONNECTION_STRING` and `VV_COSMOS_DB_ENDPOINT` from Azure Cosmos DB Emulator (You can find more instructions [here](https://learn.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21#run-on-windows)). 
5. Run the server with below command.
  `$env:NODE_TLS_REJECT_UNAUTHORIZED=0; npm run start:server`
  > Please note that `$env:NODE_TLS_REJECT_UNAUTHORIZED=0` flag is for development purpose. You can find more information [here](https://github.com/MicrosoftDocs/azure-docs/issues/57861)).
6. Run the client.
7. Click the `Click here to test API` button on the Visit page as the screenshot below:
image
8. After click the button, you can see the result as below:
  **Client**
  image
  image
  image
  **Azure Cosmos DB Emulator**
  image

# Process & policy checklist


server/src/utils/cosmosClient.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { CosmosClient, CosmosClientOptions, Database, Container, ContainerRequest } from '@azure/cosmos';
+import { DefaultAzureCredential } from '@azure/identity';
+
+const defaultAzureCredential = new DefaultAzureCredential();
+
+const cosmosClientOptions: CosmosClientOptions = {
+  endpoint: process.env.VV_COSMOS_DB_ENDPOINT!,
```
hansung-msft, line 10: [Non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) `!`

server/src/controllers/surveyResultController.ts
```diff
@@ -0,0 +1,44 @@
...
+    // Create database and container.
+    await createDatabase();
+    await createContainer();
+
+    // Store survey result.
+    await cosmosClient.database(virtualVisitsDatabaseName).container(surveyContainerName).items.upsert(survey);
+
+    res.json('Store successfully.');
+  } catch (error) {
+    return next(error);
```
hansung-msft, line 42: [Express asynchronous error handling](https://expressjs.com/en/guide/error-handling.html#:~:text=You%20must%20catch%20errors%20that%20occur%20in%20asynchronous%20code%20invoked%20by%20route%20handlers%20or%20middleware%20and%20pass%20them%20to%20Express%20for%20processing.%20For%20example%3A)

server/src/controllers/surveyResultController.ts
```diff
@@ -0,0 +1,44 @@
...
+  virtualVisitsDatabaseName,
+  surveyContainerName,
+  cosmosClient,
+  createDatabase,
+  createContainer
+} from '../utils/cosmosClient';
+
+export const storeSurveyResult = async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+  try {
+    const { body: requestData } = req;
```
hansung-msft, line 17: [Destructuring assignment - Assigning to new variable names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#:~:text=Copy%20to%20Clipboard-,Assigning%20to%20new%20variable%20names,-A%20property%20can)

server/src/app.ts
```diff
@@ -18,13 +19,23 @@ app.disable('x-powered-by');
 app.use((req, res, next) => {
   res.setHeader('X-XSS-Protection', '1; mode=block');
   res.setHeader('X-Content-Type-Options', 'nosniff');
+  res.setHeader('Access-Control-Allow-Origin', '*');
+  res.setHeader('Access-Control-Allow-Methods', 'DELETE, POST, GET, OPTIONS');
```
rvenugopal-msft, line 23: Do we have any methods with Delete and Options yet? Do we really need to set those?

server/src/controllers/surveyResultController.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const inputData: SurveyResultRequestModel = {
+      ...requestData,
+      dateTime: new Date()
+    };
+
+    const survey = new SurveyResultRequest(inputData);
+
+    // Create database and container.
+    await createDatabase();
```
rvenugopal-msft, line 34: Could this be abstracted within CosmosClient and initialize method could take care this once?

server/src/utils/cosmosClient.ts
```diff
@@ -0,0 +1,47 @@
...
+  aadCredentials: defaultAzureCredential
+};
+
+const initializeCosmosClient = () => {
+  if (process.env.VV_COSMOS_DB_CONNECTION_STRING) {
+    return new CosmosClient(process.env.VV_COSMOS_DB_CONNECTION_STRING);
```
rvenugopal-msft, line 16: This instance should be encapsulated inside this class. It should not be exposed outside

server/src/controllers/surveyResultController.ts
```diff
@@ -0,0 +1,44 @@
...
+    await createContainer();
+
+    // Store survey result.
+    await cosmosClient.database(virtualVisitsDatabaseName).container(surveyContainerName).items.upsert(survey);
```
rvenugopal-msft, line 38: could you please abstract the database details within the cosmosClient. IMO the database details should not bleed out into other classes. 

That should look something like , so using the instance of /util/CosmosClient which you have defined. 
cosmosClient.upsert(surveyContainerName, survey)

server/src/app.ts
```diff
@@ -47,6 +58,8 @@ const identityClient =
 app.get('/api/config', configController(config));
 app.get('/api/token', tokenController(identityClient, config));
 
+app.post('/api/createSurveyResult', storeSurveyResult);
+
```
rvenugopal-msft, line 62: should util\CosmosClient instance be created here once and injected into controller classes wherever needed?

server/src/app.ts
hansung-msft: Without this options, I had CORS issue so I am planning to manually deploy one with options and another without options to see if I get a CORS issue.

hansung-msft: I added this to resolve the CORS issue. If we have to put this, I will delete `DELETE` method since we need other methods.

server/src/utils/cosmosClient.ts
hansung-msft: Not sure if this is exposed but I can change CosmosClient to be class.

server/src/app.ts
hansung-msft: This is interesting idea. I am thinking that this layer is more router layer that not sure if we want to pass `cosmosClient` instance and pass it to certain controller methods

server/src/utils/cosmosClient.ts
```diff
@@ -0,0 +1,47 @@
...
+  } else {
+    return new CosmosClient(cosmosClientOptions);
+  }
+};
+
+export const cosmosClient = initializeCosmosClient();
+
+export const virtualVisitsDatabaseName = process.env.VV_COSMOS_DB_NAME!;
+export const surveyContainerName = 'Surveys';
```
rvenugopal-msft, line 25: Should we have some uniqueness around `sessionId` and `userId` in database. I believe by blindly persisting what we are getting through Api request, we are opening ourselves to duplicate records, if an end user somehow submits multiple surveys for a session.

This could accordingly be checked in this class before persisting if the record exists by the sessionId and userId. If exists just return a success.

server/src/app.ts
rvenugopal-msft: This class is the middleware layer. The middleware layer usually takes care of creating the instances which are supposed to be kept alive for the lifetime of the application. And then takes care of passing these to the classes where required.

You would want the db layer to be instantiated only once in the application as it's an expensive operation.

server/src/utils/cosmosClient.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation.
```
rvenugopal-msft, line 1: instead of having /utils , could be define the class under /db

server/src/models/surveyResultRequest.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { SurveyResultRequestModel } from '../interfaces/surveyModel';
+
+class SurveyResultRequest {
```
rvenugopal-msft, line 6: if this is an entity to be persisted within database, then could you move this class as under src/db/models . Similar comment for CosmosClient class as well.

server/src/interfaces/surveyModel.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface SurveyResultRequestModel {
```
rvenugopal-msft, line 4: No need of moving these model classes from under the src/models, I would rather suggest to create the src/db/models to handle the db entity.

server/src/utils/cosmosClient.ts
hansung-msft: Hmm I am not sure how to make it `sessionId` and `userId` unique to prevent the duplicate .. I will need help on this.

server/src/controllers/surveyController.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (!response) errors.push('response is missing');
+      return res.status(400).send({ errors });
+    }
+
+    const inputData: SurveyResultRequestModel = {
+      ...requestData,
+      createdOn: new Date()
+    };
+
+    await surveyService.init();
```
rvenugopal-msft, line 34: Can this be instantiated only once. You are creating service instance with handling of every request and service instance is also creating dB instance in every call, that seems like an expensive operation.

server/src/utils/cosmosClient.test.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
```
rvenugopal-msft, line 3: can this be src/db/cosmosClient.test.ts

server/src/services/surveyService.ts
```diff
@@ -0,0 +1,33 @@
...
+
+class SurveyService {
+  private cosmosClient: CosmosClient;
+
+  constructor(cosmosClient: CosmosClient) {
+    this.cosmosClient = cosmosClient;
+  }
+
+  async init(): Promise<void> {
+    await this.cosmosClient.createDatabase();
```
rvenugopal-msft, line 24: I believe line 24 does not need to be invoked here. It still feels cosmosClient logic. Its not required here. Db should be instantiated as part of cosmosClient contructor.

server/src/controllers/surveyController.ts
hansung-msft: That make sense however, I wasn't able to run `surveyService.init()` in `app.ts` since the `init()` is asynchronous action that is why I decided to run it under the controller level.

server/src/utils/cosmosClient.test.ts
hansung-msft: The tests are not pushed since the implementation has been shifted that I decided to fix the test after agreement.

PR #96, opened 2022-09-27
Hansung/one question poll api

hansung-msft:
# What

- One question poll API has been added.
- Test cases has been added.
- Removed the general errors. It is now response with proper error message.

# How Tested

```
specs:
node -v: 16.15.0
npm -v: 8.5.5
```
1. [Download Azure Cosmos DB Emulator](https://learn.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21) on the local.
2. Run `npm i` under `/server` directory.
3. Set `communicationServicesConnectionString` in `defaultConfig.json` file.
4. Add following environment variables:  
    ```
      VV_COSMOS_DB_CONNECTION_STRING: <your_db_connection_string>
      VV_COSMOS_DB_ENDPOINT: <your_db_endpoint>
      VV_COSMOS_DB_NAME: <your_db_name>
    ```
    You can find `VV_COSMOS_DB_CONNECTION_STRING` and `VV_COSMOS_DB_ENDPOINT` from Azure Cosmos DB Emulator (You can find more instructions [here](https://learn.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21#run-on-windows)). 
5. Run the server with below command.
  `$env:NODE_TLS_REJECT_UNAUTHORIZED=0; npm run start:server`
  > Please note that `$env:NODE_TLS_REJECT_UNAUTHORIZED=0` flag is for development purpose. You can find more information [here](https://github.com/MicrosoftDocs/azure-docs/issues/57861)).
6. You can copy the code that create button and send API from the client side from [here](https://github.com/Azure-Samples/communication-services-virtual-visits-js/pull/94/commits/3d938a0ff96fb73c7b95cb1f880b3c20e5e95b9b). 
> If you decided to use other API tool such as Postman, you don't have to follow the below steps.
7. Once you copied the code from the link above, run the client.
8. Click the `Click here to test API` button on the Visit page as the screenshot below:
image
9. After click the button, you can see the result as below:
  **Client**
  image
  image
  image
  **Azure Cosmos DB Emulator**
  image

# Process & policy checklist


server/src/databases/cosmosClient.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  afterEach(() => {
+    delete process.env.VV_COSMOS_DB_CONNECTION_STRING;
+  });
+
+  test('Initialize cosmosClient when VV_COSMOS_DB_CONNECTION_STRING present.', async () => {
+    const expectedEndpoint = 'https://testinghost/';
+    const expectedAccountKey = 'testAccountKey';
+    process.env.VV_COSMOS_DB_CONNECTION_STRING = `AccountEndpoint=${expectedEndpoint};AccountKey=${expectedAccountKey}`;
+
+    const config = getServerConfig();
```
ralphgabrinao, line 18: I think we can just mock the config and not call external functions so that the unit tests are isolated.

server/src/utils/getConfig.ts
```diff
@@ -86,7 +89,10 @@ export const getServerConfig = (): ServerConfigModel => {
     waitingTitle: process.env[VV_WAITING_TITLE_ENV_NAME] ?? defaultConfig.waitingTitle,
     waitingSubtitle: process.env[VV_WAITING_SUBTITLE_ENV_NAME] ?? defaultConfig.waitingSubtitle,
     logoUrl: process.env[VV_LOGO_URL_ENV_NAME] ?? defaultConfig.logoUrl,
-    postCall: getPostCallConfig(defaultConfig)
+    postCall: getPostCallConfig(defaultConfig),
+    cosmosDBConnectionString: process.env[VV_COSMOS_DB_CONNECTION_STRING],
+    cosmosDBEndpoint: process.env[VV_COSMOS_DB_ENDPOINT],
+    cosmosDBName: process.env[VV_COSMOS_DB_NAME]
```
ralphgabrinao, line 95: The cosmosDB configs should also be fetched from the defaultConfig

`process.env[VV_COSMOS_DB_NAME] ?? defaultConfig.cosmosDBName`

server/src/controllers/configController.test.ts
```diff
@@ -60,7 +66,11 @@ describe('configController', () => {
       colorPalette: '',
       waitingTitle: '',
       waitingSubtitle: '',
-      logoUrl: ''
+      logoUrl: '',
+      postCall: {},
```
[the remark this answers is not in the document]
hansung-msft: Since `postCall` is optional property, I will remove it from the test.

server/src/models/configModel.ts
```diff
@@ -35,6 +35,9 @@ export interface ServerConfigModel {
   waitingSubtitle: string;
   logoUrl: string;
   postCall?: PostCallConfig;
+  cosmosDBConnectionString: string | undefined;
+  cosmosDBEndpoint;
```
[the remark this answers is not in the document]
hansung-msft: Good catch! I will add `string` type there.

server/src/controllers/configController.test.ts
```diff
@@ -48,6 +51,9 @@ describe('configController', () => {
     expect(resp.lastJson.waitingTitle).toBe('str5');
     expect(resp.lastJson.waitingSubtitle).toBe('str6');
     expect(resp.lastJson.logoUrl).toBe('str7');
+    expect(resp.lastJson).not.toHaveProperty('cosmosDBConnectionString');
+    expect(resp.lastJson).not.toHaveProperty('cosmosDBEndpoint');
+    expect(resp.lastJson).not.toHaveProperty('cosmosDBName');
```
ralphgabrinao, line 56: I think these assertions are more suited for the `getClientConfig` tests, since `getClientConfig` is being invoked by configController.

server/src/models/configModel.ts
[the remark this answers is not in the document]
ralphgabrinao: I think the cosmosDB properties should all be optional. For example, if I don't need Cosmos DB (i.e. if I choose post call to be 'None', 'MSForms', or 'Custom'), then I should be able to ignore these Cosmos DB properties.

Then, we should properly handle the case where:
- Post Call is set to OneQuestionPoll
- But can't initialize Cosmos (i.e. because connection string is missing, or endpoint is missing, or db name is missing, etc.)
In that case we should return an error.

Conversely, if Post Call is **not** OneQuestionPoll, then it should be OK for the Cosmos properties to be undefined or empty strings.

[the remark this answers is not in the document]
ralphgabrinao: Also consider making an object for cosmos instead of having a flat structure. Something like:

```
{
    ...
    cosmosDb?: {
        connectionString?: string;
        endpoint? string;
        dbName: string
    }
}
```

server/src/controllers/surveyController.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const inputData: SurveyResultRequestModel = {
+      ...requestData,
+      createdOn: new Date()
+    };
+
+    await surveyDBHandler.init();
+    await surveyDBHandler.saveSurveyResult(inputData);
+
+    res.json('Store successfully.');
```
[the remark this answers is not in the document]
hansung-msft: I updated the code and now it doesn't return anything but the status code is 201.

server/src/constants.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+export const VV_COSMOS_DB_CONNECTION_STRING = 'VV_COSMOS_DB_CONNECTION_STRING';
+export const VV_COSMOS_DB_ENDPOINT = 'VV_COSMOS_DB_ENDPOINT';
+export const VV_COSMOS_DB_NAME = 'VV_COSMOS_DB_NAME';
```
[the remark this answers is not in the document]
hansung-msft: Yes we can put the db configs at the bottom of the list.

server/src/utils/getConfig.ts
hansung-msft: If we planned to include the db configs in `defaultConfig`, then we can grab it from it. With current implementation, nothing was from the `defaultConfig`.

server/src/databases/cosmosClient.ts
```diff
@@ -0,0 +1,67 @@
...
+  private _database: Database | undefined;
+  private _container: Container | undefined;
+
+  constructor({ cosmosDBConnectionString, cosmosDBEndpoint, cosmosDBName }: ServerConfigModel) {
+    const cosmosClientOptions: CosmosClientOptions = {
+      endpoint: cosmosDBEndpoint,
+      aadCredentials: new DefaultAzureCredential()
+    };
+
+    if (cosmosDBConnectionString) {
```
ralphgabrinao, line 25: I think we should also check for emptiness here, to handle the case where conn string is empty but endpoint is defined. That is to say, default to using the endpoint if connstring is not valid.
`if (cosmosDBConnectionString && cosmosDBConnectionString.length > 0)`

server/src/controllers/surveyController.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      return res.status(400).send({ errors });
+    }
+
+    const inputData: SurveyResultRequestModel = {
+      ...requestData,
+      createdOn: new Date()
+    };
+
+    await surveyDBHandler.init();
```
mbakalov, line 35: Shouldn't `init` on every controller request. Do this once on the app startup instead.

server/src/databases/cosmosClient.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  async createContainer(containerRequest: ContainerRequest): Promise<void> {
+    if (this._container === undefined) {
+      const { container } = await this.database(this._cosmosDBName).containers.createIfNotExists(containerRequest);
+
+      this._container = container;
+    }
+  }
+
+  async upsert(containerId: string, item: any): Promise<void> {
```
mbakalov, line 54: Should strongly-type the `item`

server/src/controllers/surveyController.ts
hansung-msft: I tried to `init` in `app.ts` but `init` is asynchronous function that I wasn't able to run it on `app.ts`. Do you know the way I can run asynchronous function in `app.ts` somehow?

server/src/databases/cosmosClient.ts
hansung-msft: I will replace the type `any`.

server/src/app.ts
```diff
@@ -47,6 +60,11 @@ const identityClient =
 app.get('/api/config', configController(config));
 app.get('/api/token', tokenController(identityClient, config));
 
+const cosmosClient = new CosmosClient(config);
+const surveyDBHandler = new SurveyDBHandler(cosmosClient);
+
+app.post('/api/createSurveyResult', storeSurveyResult(surveyDBHandler));
```
mbakalov, line 66: Make this `POST /api/surveyResults` maybe?

server/src/controllers/surveyController.ts
hansung-msft: Because of the above reason, inside both [`createDatabase`](https://github.com/Azure-Samples/communication-services-virtual-visits-js/pull/96/files#diff-f8bfe563ab4c43a502d9afa610dbdb917b21a0f688128c45fbfbe2825aa79427:~:text=if%20(this._database%20%3D%3D%3D%20undefined)%20%7B) and [`createContainer`](if (this._container === undefined) {) have if statement to not invoke `createIfNotExsists` method at all.

```javascript
if (this._database === undefined) {
...
}

if (this._container === undefined) {
...
}
```

server/src/app.ts
hansung-msft: Since it is post method, I think that make sense! I will update the code.

hansung-msft: What do you think about `surveyResult` since it is only handling to store one request per api call ?

hansung-msft: Actually I am going with `api/surveyResults` 😄

hansung-msft:
Added ui-change tag to trigger update snapshots.

server/src/app.test.ts
```diff
@@ -86,8 +90,21 @@ describe('errors', () => {
...
     // calling the /api/token endpoint without proper mocking throws TypeError
+    const expectedError = { error: 'client.createUserAndToken is not a function' };
     const getResponse = await request(app).get('/api/token');
     expect(getResponse.status).toEqual(500);
-    expect(getResponse.text).toEqual(JSON.stringify(ERROR_PAYLOAD_500));
+    expect(getResponse.text).toEqual(JSON.stringify(expectedError));
+  });
+
+  test('return 200 sending option request', async () => {
+    const getResponse = await request(app).options('/api/surveyReesult');
```
ralphgabrinao, line 100: Typo here in the endpoint

server/src/app.ts
```diff
@@ -56,7 +75,7 @@ app.use((err, req, res, next) => {
     return next(err);
   }
 
-  res.status(500).json(ERROR_PAYLOAD_500);
+  res.status(500).send({ error: err.message });
```
ralphgabrinao, line 78: Should we add a null check here to see if `err` is defined?

server/src/controllers/surveyController.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      callId: 'test_call_id',
+      acsUserId: 'test_acs_user_id',
+      response: true
+    };
+    const request: any = { body: inputData };
+
+    const config = getServerConfig();
+    let cosmosClient;
+
+    if (config.cosmosDb) {
```
ralphgabrinao, line 41: To keep separation of concerns and to keep the spirit of unit testing, I think we can assume here that cosmosDb is defined, and just create a mock config (no need to call `getServerConfig()` because then these become integration tests.).

IMO testing whether cosmosDb is defined or not is more suited for app.test.ts, not here. In fact you already are covering that case with this test case `check if /api/surveyResults route is open without cosmosDb configs`.

server/src/models/surveyModel.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface SurveyResultRequestModel {
+  dateTime: Date;
```
ralphgabrinao, line 5: Is this meant to be `createdOn`?

server/src/app.test.ts
```diff
@@ -86,8 +90,21 @@ describe('errors', () => {
...
+
+    expect(getResponse.status).toBe(200);
+  });
+
+  test('check if /api/surveyResults route is open without cosmosDb configs', async () => {
```
ralphgabrinao, line 105: It is not clear how the cosmosDb is set to undefined here. If I set my CosmosDb connection string env variable, will this fail?

I see there is a related test case in `surveyRoute.test.ts`, shall we move this one there as well? That way we have a way of explicitly setting `cosmosDb` to undefined.

server/src/utils/getConfig.ts
```diff
@@ -130,6 +135,36 @@ const getPostCallConfig = (defaultConfig: ServerConfigModel): PostCallConfig | u
...
+  }
+
+  if (cosmosDBEndpoint) {
+    cosmosDBConfig = {
+      ...cosmosDBConfig,
+      endpoint: cosmosDBEndpoint
+    };
+  }
+
+  return cosmosDBConfig;
```
ralphgabrinao, line 165: I think we can make this more concise:

```
const cosmosDBConfig = {
    connectionString: process.env[VV_COSMOS_DB_CONNECTION_STRING] ?? defaultConfig.cosmosDb?.connectionString,
    endpoint: process.env[VV_COSMOS_DB_ENDPOINT] ?? defaultConfig.cosmosDb?.endpoint,
    dbName: process.env[VV_COSMOS_DB_NAME] ?? (defaultConfig.cosmosDb?.dbName as string)
};

if (!(cosmosDBConfig.connectionString || cosmosDBConfig.endpoint)) {
    return undefined;
}

return cosmosDBConfig;
```

server/src/app.test.ts
```diff
@@ -6,10 +6,25 @@
 
 import request from 'supertest';
 import app from './app';
-import { ERROR_PAYLOAD_500 } from './errors';
 import fs from 'fs';
 import path from 'path';
 
+jest.mock('./utils/getConfig', () => {
```
ralphgabrinao, line 12: Is this needed and does this actually work?

server/src/app.test.ts
```diff
@@ -86,8 +104,9 @@ describe('errors', () => {
 
   test('returns a generic 500 response when an endpoints throws an error', async () => {
     // calling the /api/token endpoint without proper mocking throws TypeError
+    const expectedError = { error: 'client.createUserAndToken is not a function' };
```
ralphgabrinao, line 107: This makes it seem like we are throwing this specific error message on purpose, whereas the reason why this is the error message is because we are setting `({} as CommunicationIdentityClient)` as the identityClient if the environment is 'test'. This doesn't actually apply in a production environment. IMO we don't need to check for this specific error message.

server/src/controllers/surveyController.test.ts
```diff
@@ -0,0 +1,179 @@
...
+            return {
+              items: {
+                upsert: mockedUpsert
+              }
+            };
+          })
+        };
+      })
+    };
+    const config = getServerConfig();
```
ralphgabrinao, line 49: To make this a pure unit test, I don't think we should call `getServerConfig` or any external dependency here. Unit tests should isolate the unit under test and avoid calling dependencies that can pollute the test environment. Especially true for `getServerConfig`, which depends on the machine's environment variables and/or the defaultConfig.json. `getServerConfig` can also change under the hood which will affect this test.

Instead, we can just create the config directly:
`const config = { ... }`

server/src/databases/handlers/surveyDBHandler.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  beforeEach(() => {
+    process.env.VV_COSMOS_DB_ENDPOINT = 'https://testinghost/';
+  });
+
+  afterEach(() => {
+    delete process.env.VV_COSMOS_DB_ENDPOINT;
+  });
+
+  test('Test init', async () => {
+    const config = getServerConfig();
```
ralphgabrinao, line 17: Similar comment here about calling external dependencies in unit tests. We should create mock configs instead where we control the actual values.

server/src/models/surveyModel.ts
ralphgabrinao: [bump]

server/src/surveyRoute.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const inputData: any = {
+      sessionId: 'test_session_id',
+      callId: 'test_call_id',
+      acsUserId: 'test_acs_user_id',
+      response: true
+    };
+
+    jest.spyOn(surveyDBHandlerUtil, 'createSurveyDBHandler').mockReturnValueOnce(undefined);
+
+    const app = (await import('./app')).default;
```
ralphgabrinao, line 22: This is an interesting/clever approach. Does this still work if we move these tests to `app.test.ts`?

As much as possible I'd like to follow the "1 file has 1 test file" approach, and putting these tests in `app.test.ts` makes sense to me since `app` is the unit under test.

server/src/utils/getConfig.test.ts
```diff
@@ -288,4 +291,88 @@ describe('config', () => {
...
+    expect(config.cosmosDb).toBeUndefined();
+  });
+
+  test('getServerConfig returns cosmosDb with connection string with database name from environment variable', () => {
+    process.env.VV_COSMOS_DB_CONNECTION_STRING = 'testingConnectionString';
+    process.env.VV_COSMOS_DB_NAME = 'testingDBName';
+
+    const config = getConfig.getServerConfig();
+
+    expect(config.cosmosDb).not.toBeUndefined();
```
ralphgabrinao, line 307: `.not.toBeUndefined()` I think we can change this to `.toBeDefined()` ?

server/src/utils/getConfig.ts
ralphgabrinao: [bump]

server/src/utils/surveyDBHandlerUtil.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      },
+      cosmosDb: {
+        endpoint: 'http://testinghost.com',
+        dbName: 'testingDBName'
+      }
+    };
+
+    const surveyDBHandler = createSurveyDBHandler(config as ServerConfigModel);
+
+    expect(surveyDBHandler).not.toBeUndefined();
```
ralphgabrinao, line 55: `.toBeDefined()` here too

server/src/utils/getConfig.ts
ralphgabrinao: Reposting previous comment -

I think we can make this more concise:
```
const cosmosDBConfig = { 
    connectionString: process.env[VV_COSMOS_DB_CONNECTION_STRING] ?? defaultConfig.cosmosDb?.connectionString,
    endpoint: process.env[VV_COSMOS_DB_ENDPOINT] ?? defaultConfig.cosmosDb?.endpoint,
    dbName: process.env[VV_COSMOS_DB_NAME] ?? (defaultConfig.cosmosDb?.dbName as string)
};

if (!(cosmosDBConfig.connectionString || cosmosDBConfig.endpoint)) {
    return undefined;
}

return cosmosDBConfig;
```

server/src/app.test.ts
hansung-msft: If we think that unnecessary to check the error message, I can simply check the status code only.

server/src/controllers/surveyController.test.ts
hansung-msft: Okay that make sense. I thought since we are mocking the `cosmosClient`, the config doesn't matter what it is. But I will update on it.

server/src/app.test.ts
ralphgabrinao: That works, and we can also check the structure of the response:
```
expect(JSON.parse(getResponse.text)).toHaveProperty('error');
```

server/src/databases/handlers/surveyDBHandler.test.ts
hansung-msft: Yes I will update the test.

server/src/surveyRoute.test.ts
hansung-msft: I am not sure if this will work if I move it to `app.test.ts`. If it works, I will move it and remove the file. Otherwise, I will keep it as it is. I will update the comment.

server/src/utils/getConfig.test.ts
hansung-msft: I think either way is fine to me. I will update it to `toBeDefined()`.

server/src/utils/surveyDBHandlerUtil.test.ts
hansung-msft: This one as well!

server/src/app.ts
ralphgabrinao, line 78: Null check here for `err` ?

server/src/surveyRouteWithConfig.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    const mockedSurveyDBHandler = {
+      init: jest.fn()
+    };
+    const mockStoreSurveyResult = jest
+      .spyOn(SurveyController, 'storeSurveyResult')
+      .mockImplementationOnce(() => async () => Promise.resolve());
+
+    jest.spyOn(SurveyDBHandler, 'createSurveyDBHandler').mockReturnValueOnce(mockedSurveyDBHandler as any);
+
+    (await import('./app')).default;
```
ralphgabrinao, line 53: I see that this approach works when moved to `app.test.ts`. Should we move these tests too so that all test cases are in one place?

hansung-msft: hmm I will try. The reason I was able to merge it because it didn't require the `postCall` and `cosmosDb` configs. I will try to merge it.

server/src/models/surveyModel.ts
hansung-msft: Yea I will update it!

server/src/utils/getConfig.ts
hansung-msft: In this way, if the `endpoint` is undefined, it will result as below:
```javascript
{
  connectionString: 'asdfasfasdf',
  endpoint: undefined,
  dbName: 'testingDBName
}
```

I was trying to avoid to add field that is `undefined`.

server/src/databaseHandlers/surveyDBHandler.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    const surveyDBHandler = new SurveyDBHandler(mockedCosmosClient as any, cosmosDBConfig);
+
+    surveyDBHandler.saveSurveyResult(inputData);
+
+    expect(mockedUpsert).toHaveBeenCalled();
+  });
+});
+
+describe('Test surveyDBHandlerUtils', () => {
```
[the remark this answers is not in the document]
hansung-msft: I will update the title of the test

server/src/utils/getConfig.ts
ralphgabrinao: In that case can we just do:

```
if (cosmosDBConnectionString) {
    cosmosDBConfig.connectionString = cosmosDBConnectionString
}

if (cosmosDBEndpoint) {
    cosmosDBConfig.endpoint = cosmosDBEndpoint
}
```
?

A little nitpicky lol I am just trying to find ways to make the code a little more concise.

server/src/app.ts
hansung-msft: I will update the code.

server/src/utils/getConfig.ts
hansung-msft: That would work if we wrap it with if statement. I will update the code.

server/src/surveyRouteWithConfig.test.ts
hansung-msft: Added test in `app.ts` to testing if the route is close when there is no config for cosmosDb.

hansung-msft:
Removed the all headers and `option` checking from `app.ts` for CORS.

ralphgabrinao (dismissed):

mbakalov (dismissed):

ralphgabrinao (approved):

dailam-msft (approved):

PR #97, opened 2022-09-29
ARM changes for Cosmos DB

tsohoni-msft:
# What
- Added some new variables for CosmosDB settings.
- Creating new CosmosDB account if survey type is one question poll.
- Creating managed identity for AppService.
- Creating new SQL role definition which has permissions to access the DB account.
- Creating new SQL role assignment to assign permissions to AppService to access the DB account.

# Why
- Scaffolding for creating Cosmos DB artifacts to save one question poll responses.

# How Tested
- Tried deploying custom template with and without survey type as one question poll.

# Process & policy checklist


ralphgabrinao:
Looks good! I wonder if we should also make these edits to `editableazuredeploy.json`. Can be in a separate PR

deploy/azuredeploy.json
```diff
@@ -84,6 +89,17 @@
       },
       "properties": {}
     },
+    {
+      "type": "Microsoft.DocumentDB/databaseAccounts",
+      "apiVersion": "2021-10-15",
```
hansung-msft, line 94: Are we intentionally entered the fixed older version on here?

deploy/azuredeploy.json
```diff
@@ -61,7 +61,12 @@
     "postCallSurveyOneQuestionPollPrompt": "",
     "postCallSurveyOneQuestionPollType": "",
     "postCallSurveyOneQuestionPollAnswerPlaceholder": "",
-    "postCallSurveyOneQuestionPollSaveButtonText": ""
+    "postCallSurveyOneQuestionPollSaveButtonText": "",
+    "cosmosDbAccountName": "[concat('db-account-', parameters('appServiceName'))]",
+    "roleDefinitionName": "[concat('sql-role-definition-', parameters('appServiceName'))]",
+    "cosmosDbName": "",
```
ralphgabrinao, line 67: We should provide a default value here since it is not a parameter/ it can easily be missed. I was able to provision the resources successfully but realized afterwards in the AppService's config that the db name was empty. I feel like that will be a common mistake.

ralphgabrinao, line 67: I think we should provide a default value here (other variables also have a default value). Since this is not a parameter, it is easily missed and while the resources will be successfully created, the db name will be left empty in the App Service's environment variables.

deploy/azuredeploy.json
tsohoni-msft: No, that was an oversight. Updated to latest api version for all the newly added resources.

deploy/azuredeploy.json
tsohoni-msft: Agreed, added a default value.

ralphgabrinao (approved):

ralphgabrinao (approved):

allenplusplus:
> Looks good! I wonder if we should also make these edits to `editableazuredeploy.json`. Can be in a separate PR

+ 1 to Ralph. I think we need a task to fix up editableazuredeploy.json. It may be broken now since it is lacking all of the post call changes while it is referencing the latest sample code.

allenplusplus (approved):

PR #98, opened 2022-10-04
1 Question Poll implementation

dailam-msft:
What
1)UI rendering for postcall
2)saving to cosmos db

Why
PostCall v2

How Tested
locally - mock post call object in defaultCOnfig.json - 
1) Change pollType - likeOrDislike / rating / text
"postCall": {
    "survey": {
      "type": "onequestionpoll",
      "options": {
        "title": "Tell us how we did",
        "prompt": "How satisfied are you with this virtual appointment's audio and video quality?",
        "pollType": "text",
        "answerPlaceholder": "",
        "saveButtonText": "Continue"
      }
    }
  },,
  "cosmosDb": {
    "connectionString" :"<your endpoint>",
    "dbName" :"<your db name>"
  }
image
image
image


client/src/components/postcall/Survey.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const Survey: React.FunctionComponent<SurveyProps> = (props: SurveyProps) => {
+  const surveyType = props.postCall.survey?.type;
+  let postcallSurveyUrl = '';
+  if (surveyType === 'msforms') {
+    const options: MSFormsSurveyOptions = props.postCall.survey?.options as MSFormsSurveyOptions;
+    postcallSurveyUrl = options.surveyUrl;
+  } else if (surveyType === 'custom') {
+    const options: CustomSurveyOptions = props.postCall.survey?.options as CustomSurveyOptions;
+    postcallSurveyUrl = options.surveyUrl;
+  } else {
```
ralphgabrinao, line 32: This `else` block is not necessary

client/src/components/postcall/RejoinLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Link } from '@fluentui/react';
+
+export interface RejoinLinkProps {
+  onRejoinCall: () => void;
+}
+
+export const RejoinLink: React.FunctionComponent<RejoinLinkProps> = (props: RejoinLinkProps) => {
+  return (
+    <Link
+      onClick={() => {
```
ralphgabrinao, line 14: Does this work?

```
onClick={props.onRejoinCall}
```

client/src/components/postcall/Survey.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <RejoinLink
+            onRejoinCall={() => {
+              void props.onRejoinCall();
+            }}
+          ></RejoinLink>
+        </Stack>
+      </Stack>
+    );
+  } else if (surveyType === 'onequestionpoll') {
+    const oneQuestionPollInfo: OneQuestionPollOptions = props.postCall.survey.options as OneQuestionPollOptions;
```
ralphgabrinao, line 49: Can we make the name consistent with the type?

`const oneQuestionPollOptions: OneQuestionPollOptions`

client/src/components/postcall/PostCallOneQuestionPoll.tsx
```diff
@@ -0,0 +1,32 @@
...
+) => {
+  return (
+    <Stack horizontalAlign="center" verticalAlign="center">
+      <Text styles={pollTitleStyle}>{props.oneQuestionPollInfo.title}</Text>
+      <Text styles={pollPromptStyle}>{props.oneQuestionPollInfo.prompt}</Text>
+      <OneQuestionPollInput
+        pollType={props.oneQuestionPollInfo.pollType}
+        textInputPlaceholder={props.oneQuestionPollInfo.answerPlaceholder}
+      />
+      <PrimaryButton tabIndex={-1} width="20rem" height="2.5rem">
```
hansung-msft, line 25: We might need `styles` props to change the styles on the button

client/src/components/postcall/PostCallOneQuestionPoll.tsx
```diff
@@ -0,0 +1,48 @@
...
+        onClick={() =>
+          submitSurveyResponse(
+            props.acsUserId,
+            pollResponse !== undefined ? pollResponse : undefined,
+            props.callId ? props.callId : undefined
+          )
+        }
+      >
+        <Stack horizontal verticalAlign="center">
+          <span>{props.oneQuestionPollInfo.saveButtonText}</span>
```
ralphgabrinao, line 43: To avoid code repetition, we can extract the individual variables from `props.oneQuestionPollInfo` 

```
const { title, prompt, pollType, answerPlaceholder, saveButtonText } = props.oneQuestionPollInfo;
```

client/src/components/postcall/PostCallOneQuestionPoll.tsx
```diff
@@ -0,0 +1,48 @@
...
+        textInputPlaceholder={props.oneQuestionPollInfo.answerPlaceholder}
+        setPollResponse={setPollResponse}
+      />
+      <PrimaryButton
+        styles={surveySubmitButtonStyles}
+        onClick={() =>
+          submitSurveyResponse(
+            props.acsUserId,
+            pollResponse !== undefined ? pollResponse : undefined,
+            props.callId ? props.callId : undefined
```
ralphgabrinao, line 38: Are the null checks necessary here? I believe this is logically equivalent to:

```
submitSurveyResponse(props.acsUserId, pollResponse, props.callId);
```

client/src/components/postcall/OneQuestionPollInput.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const onQuestionPollInput = await mount<OneQuestionPollInputProps>(
+      <OneQuestionPollInput
+        pollType={mockPollType}
+        textInputPlaceholder={mockTextInputPlaceholder}
+        setPollResponse={mockSetPollResponse}
+      />
+    );
+    const rating = onQuestionPollInput.find(Rating);
+    expect(rating.length).toBe(1);
+    rating.simulate('change');
```
ralphgabrinao, line 47: Is it possible to provide an actual value here so we can assert that `mockSetPollResponse` was called with that value?

client/src/components/postcall/OneQuestionPollInput.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+        setPollResponse={mockSetPollResponse}
+      />
+    );
+    const textField = onQuestionPollInput.find(TextField);
+    expect(textField.length).toBe(1);
+    textField
+      .first()
+      .find('textarea')
+      .simulate('change', { target: { value: 'Changed' } });
+    expect(mockSetPollResponse).toHaveBeenCalled();
```
ralphgabrinao, line 69: `.toBeCalledWith('Changed')`

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: prefer to leave this as it is, since when users read through the component it is clear where these values are obtained from

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: Yes they are necessary else the assignment into submitSurveyResponse fails/throws errors

ralphgabrinao: `assignment into submitSurveyResponse` Can you clarify, do you mean the API call fails?

ralphgabrinao: There's no compiler errors so I don't see a need for the null checks. If there are errors downstream then that business logic must be added there.

2022-10-12 16_18_29-PostCallOneQuestionPoll tsx - postcall-v2 - Visual Studio Code

client/src/components/postcall/RejoinLink.tsx
dailam-msft: Thanks, this works, have mad this change!

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: Have made this change!Thanks!

client/src/components/postcall/Survey.tsx
dailam-msft: have made this change

client/src/utils/PostCallUtil.ts
```diff
@@ -0,0 +1,31 @@
...
+      body: JSON.stringify({
+        callId: callId,
+        acsUserId: acsUserId,
+        response: pollResponse
+      })
+    });
+    if (res.status === 200) {
+      window.location.replace('/book');
+    } else {
+      window.location.replace('/book');
```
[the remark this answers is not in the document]
dailam-msft: This is true, looks cleaner , thanks! Have made this change

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: Agghhh... not sure what happened, was getting some errors before which I dont remember after the dozen other changes so far.. but this works fine, have made this change!

client/src/components/postcall/OneQuestionPollInput.test.tsx
dailam-msft: Added this

client/src/components/postcall/OneQuestionPollInput.test.tsx
dailam-msft: Nope , tried various combinations of click and rating and value and that doesnt seem to do the trick.

ralphgabrinao: How come you were able to do it here? https://github.com/Azure-Samples/communication-services-virtual-visits-js/pull/98/files#diff-c9dc5310ac40d64d672e151215e4fb7625f913dbc00ad6a070bccff186a3cdbbR68

client/src/components/postcall/OneQuestionPollInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+        resizable={false}
+        width="100%"
+        onChange={(_event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) =>
+          props.setPollResponse(newValue)
+        }
+      />
+    );
+  }
+
+  return <></>;
```
hansung-msft, line 107: Now I am thinking about it, should we show some kind of error saying the poll type is not correct? what do you think about it?

hansung-msft:
Also, would you be able to update the screenshots in the description please?

client/src/components/MeetingExperience.test.tsx
```diff
@@ -161,74 +161,96 @@ describe('MeetingExperience', () => {
     expect(callWithChatComposites.length).toBe(1);
   });
 
-  it('should render Survey component when postcall is defined and valid', async () => {
-    const setRenderPostCallMock = jest.fn();
-    const useStateMock: any = (_: any) => [true, setRenderPostCallMock];
-    jest.spyOn(React, 'useState').mockImplementation(useStateMock);
+  // it('should render Survey component when postcall is defined and valid', async () => {
```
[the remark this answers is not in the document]
ralphgabrinao: We should at least bring this test back: `should not render Survey component when postcall is undefined`. Then we can separately address the other case and add a test later for the case where postcall is defined and valid.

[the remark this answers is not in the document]
dailam-msft: Setting useState to true makes callWithChatApapdter get set to "true" too (since it is a state value) . if i remember right from my trial and error, the test will break in MeetingExperience.tsx where callWithChatAdapter is accessed to get the callID. testing state is considered fragile.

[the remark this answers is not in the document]
dailam-msft: https://stackoverflow.com/questions/57692521/how-to-mock-the-usestate-hook-implementation-so-that-it-actually-changes-the-sta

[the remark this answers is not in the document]
hansung-msft: I wrote some test that mocks `@azure/communication-react` package that we can control (mock) props to `MeetingExperience` component to test.

We no longer need to mock `useState` anymore since it is bad practice and it is not possible to test component that has more than one state.

Please take a look [this file](https://github.com/Azure-Samples/communication-services-virtual-visits-js/blob/4e6415473b7ad13926ec274d6d3038b9622362e5/client/src/components/MeetingExperience.test.tsx)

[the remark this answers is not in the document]
dailam-msft: that worked! nice find, please add any reference links you have on how you landed on this approach. I made some tweaks to also include callWithChatComposite check when survey is not rendered

client/src/components/postcall/OneQuestionPollInput.test.tsx
dailam-msft: Rating in this instance is a bunch of 5 different rating buttons.. it will involve finding a specific one and triggering the right event. Triggering change and click for the GROUP level element as well as the first rating button did not work .

As of now test coverage for this file is at 91.89 % . I think we can leave this

client/src/components/postcall/OneQuestionPollInput.tsx
dailam-msft: We could. Will run it by JJ and Jessica and create a work item for it.

client/src/components/MeetingExperience.test.tsx
```diff
@@ -214,21 +226,15 @@ describe('MeetingExperience', () => {
...
     meetingExperience.update();
 
-    expect(setRenderPostCallMock).toBeCalled();
     const survey = meetingExperience.find(Survey);
-    expect(survey.length).toBe(0);
-    const callWithChatComposites = meetingExperience.find(CallWithChatComposite);
-    expect(callWithChatComposites.length).toBe(1);
-    const parentDiv = callWithChatComposites.parent();
-    expect(parentDiv.props().style.display).toBe('flex');
+    expect(survey.length).toBe(1);
```
ralphgabrinao, line 238: Let's bring back the assertions on CallWithChatComposite to truly test that only Survey is visible:

```
    const callWithChatComposites = meetingExperience.find(CallWithChatComposite);
    expect(callWithChatComposites.length).toBe(1);
    const parentDiv = callWithChatComposites.parent();
    expect(parentDiv.props().style.display).toBe('none');
```

client/src/components/postcall/Survey.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const options: CustomSurveyOptions = props.postCall.survey?.options as CustomSurveyOptions;
+    postcallSurveyUrl = options.surveyUrl;
+  }
+
+  if (surveyType === 'msforms' || surveyType === 'custom') {
+    return (
+      <Stack styles={surveyStyle}>
+        <iframe title={SURVEY} style={surveyIframeStyle} src={postcallSurveyUrl} scrolling="yes"></iframe>
+        <Stack horizontalAlign="center" verticalAlign="center" styles={rejoinLinkStyle}>
+          <RejoinLink
```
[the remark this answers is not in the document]
hansung-msft: I see the refactoring of `<Survey />` component but `<RejoinLink />` is not applied.. is there specific reason for that?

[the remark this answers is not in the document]
dailam-msft: I think I missed this.. Will add this . Thanks!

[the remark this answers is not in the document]
dailam-msft: done

client/src/components/MeetingExperience.test.tsx
dailam-msft: done

ralphgabrinao (approved):

hansung-msft (approved):

dailam-msft:
Merging all changes for One Question poll implementation

PR #100, opened 2022-10-04
Removed padding,avoids new tab on small screens

dailam-msft:
# What
Removed padding to avoid opening a new tab when postcall survey is loaded on smaller screens

# Why
Comments from UX review

# How Tested
Mobile and Dektop - 
image

image


tejaswip-msft (approved):

tejaswip-msft (commented):
I see that for small screens , it shows MS forms with "Fill the form" and clicking on that leads to open it in a new tab. Is this expected behaviour?
image

hansung-msft (approved):

PR #102, opened 2022-10-07
Added input validator for surveyResultRequest.

hansung-msft:
# What

- Validation for survey result request has been added.
  - Checks if the field present.
  - If the field present, it checks each field type.
 
# How Tested

- You can test it by simply using Postman to send API with invalid payload.

# Process & policy checklist

### DIFF
--- a/server/src/controllers/surveyController.ts
+++ b/server/src/controllers/surveyController.ts
@@ -3,6 +3,7 @@
 
 import * as express from 'express';
 import SurveyDBHandler from '../databaseHandlers/surveyDBHandler';
+import { surveyResultRequestValidator } from '../utils/validators';
 import { SurveyResultRequestModel } from '../models/surveyModel';
 
 export const storeSurveyResult = (surveyDBHandler: SurveyDBHandler) => async (
@@ -12,17 +13,10 @@ export const storeSurveyResult = (surveyDBHandler: SurveyDBHandler) => async (
 ): Promise<any> => {
   try {
     const { body: requestData } = req;
-    const { callId, acsUserId, response } = requestData;
 
     // Validation.
-    // If any one of field does not exists in the payload the validation failed.
-    if (!callId || !acsUserId || response === undefined) {
-      const errors: string[] = [];
-
-      if (!callId) errors.push('callId is missing');
-      if (!acsUserId) errors.push('acsUserId is missing');
-      if (response === undefined) errors.push('response is missing');
-
+    const errors = surveyResultRequestValidator(requestData);
+    if (errors.length > 0) {
       return res.status(400).send({ errors });
     }
 
--- a/server/src/controllers/surveyController.ts
+++ b/server/src/controllers/surveyController.ts
@@ -15,7 +15,8 @@ export const storeSurveyResult = (surveyDBHandler: SurveyDBHandler) => async (
     const { body: requestData } = req;
 
     // Validation.
-    const errors = surveyResultRequestValidator(requestData);
+    const errors = await surveyResultRequestValidator(requestData, surveyDBHandler);
+
     if (errors.length > 0) {
       return res.status(400).send({ errors });
     }
--- a/server/src/utils/validators.test.ts
+++ b/server/src/utils/validators.test.ts
@@ -1,7 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+import SurveyDBHandler from '../databaseHandlers/surveyDBHandler';
 import { surveyResultRequestValidator } from './validators';
+
+const cosmosDBConfig = {
+  endpoint: 'https://example.org',
+  dbName: 'testingDbName'
+};
 
 describe('validators test', () => {
   describe('testing surveyResultRequestValidator', () => {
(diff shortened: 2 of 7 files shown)


server/src/utils/validators.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { SurveyResultRequest } from '../models/surveyModel';
+
+export const surveyResultRequestValidator = (requestData: SurveyResultRequest): string[] => {
+  const { callId, acsUserId, response } = requestData;
+  const errors: string[] = [];
+  const responseAllowedtypes = ['boolean', 'number', 'string'];
+
+  // Check fields are undefined
+  if (!callId) errors.push('callId is missing');
```
rvenugopal-msft, line 12: instead of "is missing" can it be replaced with "must be present"

server/src/controllers/surveyController.ts
```diff
@@ -12,17 +13,10 @@ export const storeSurveyResult = (surveyDBHandler: SurveyDBHandler) => async (
...
     // Validation.
-    // If any one of field does not exists in the payload the validation failed.
-    if (!callId || !acsUserId || response === undefined) {
-      const errors: string[] = [];
-
-      if (!callId) errors.push('callId is missing');
-      if (!acsUserId) errors.push('acsUserId is missing');
-      if (response === undefined) errors.push('response is missing');
-
+    const errors = surveyResultRequestValidator(requestData);
```
rvenugopal-msft, line 18: should you also validate for the duplicate record, ignore if already saved?

server/src/utils/validators.ts
hansung-msft: That sounds better. I will update the code!

server/src/controllers/surveyController.ts
hansung-msft: hmm in that case, how can we determine duplicate record? Should I check `callId` ? or should I check all of the field?

hansung-msft: Added the logic to check if the survey is already exists or not.

server/src/utils/validators.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // Check if each field types are correct
+  if (callId && typeof callId !== 'string') errors.push('callId type must be string');
+  if (acsUserId && typeof acsUserId !== 'string') errors.push('acsUserId type must be string');
+  if (response !== undefined && !responseAllowedtypes.includes(typeof response))
+    errors.push('response type must be one of boolean, string, number');
+
+  const queryResults = await surveyDBHandler.querySurvey(callId, acsUserId);
+
+  if (queryResults.length > 0) errors.push('you can only submit survey once');
```
ralphgabrinao, line 28: I think we can be more specific here why the request was denied:

`Response has already been recorded for call id <callId> and user id <userId>`

hansung-msft: That sound better. I will update it!

ralphgabrinao (dismissed):

server/src/utils/validators.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const surveyResultRequestValidator = async (
+  requestData: SurveyResultRequest,
+  surveyDBHandler: SurveyDBHandler
+): Promise<string[]> => {
+  const { callId, acsUserId, response } = requestData;
+  const errors: string[] = [];
+  const responseAllowedtypes = ['boolean', 'number', 'string'];
+
+  // Check fields are undefined
```
tejaswip-msft, line 15: tiny nitpick: this can be changed to "check if fields are undefined"

server/src/utils/validators.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!callId) errors.push('callId must be present');
+  if (!acsUserId) errors.push('acsUserId must be present');
+  if (response === undefined) errors.push('response must be present');
+
+  // Check if each field types are correct
+  if (callId && typeof callId !== 'string') errors.push('callId type must be string');
```
tejaswip-msft, line 21: is checking if callId is defined necessary here? we are already doing that before this step right

server/src/utils/validators.test.ts
```diff
@@ -0,0 +1,122 @@
...
+import { surveyResultRequestValidator } from './validators';
+
+const cosmosDBConfig = {
+  endpoint: 'https://example.org',
+  dbName: 'testingDbName'
+};
+
+describe('validators test', () => {
+  describe('testing surveyResultRequestValidator', () => {
+    describe('testing when fields is not present', () => {
```
tejaswip-msft, line 14: tiny nitpick, "testing when fields are not present"

server/src/utils/validators.ts
hansung-msft: I think so. This is because if we don't check if `callId` existence, the `callId type must be string` error will include also and I think it is strange since the `callId` is not even present.

server/src/utils/validators.ts
hansung-msft: I will update it!

server/src/utils/validators.test.ts
hansung-msft: I will update it!

tejaswip-msft (approved):

ralphgabrinao (approved):

PR #109, opened 2022-10-14
Removed querySurvey and activate the unique key constraints in Cosmos…

hansung-msft:
# What

- As we discussed from the demo, instead of manually querying database to see if the survey result has been saved, I was able to activate the unique key constraints in Cosmos DB.
- In this PR, it checks `callId` and we can't insert any data with same `callId`.

**Update**
- ~~New logic has been added the the middleware to handle errors from `/api/surveyResults` route. As the screenshot below in _Result_ section, now it is return 409 for conflict data.~~
- Revert the last changes. 

# Why

- In this way, we don't have to worry about racing condition, and reduced unnecessary interaction with DB.

# How Tested

1. Modify the either environment variable or `defaultConfig` to use cosmos DB.
2. You can connect to either Cosmos DB Emulator or Portal Cosmos DB.
3. Use Postman to send test api.

# Result
> I was using Cosmos DB Emulator and used Postman during testing.
1. Start the Cosmos DB Emulator and server.
  **Cosmos DB:**
  image
  Currently, it is empty since just start the server. You can see the container and database has been created.

2. Submit the first survey result. I used the Postman as below and sent it.
  **Postman:**
  image
  As you can see above, received 200 status code and successfully stored in Cosmos DB as below:
  **Cosmos DB:**
  image

3. Send another request with modified `acsUserID`. Request fail with 500 status code with detail message.
  **Postman:**
  image
  And no changes in Cosmos DB.
  **Cosmos DB:**
  image

# Unit test result
image

# Process & policy checklist


server/src/databaseHandlers/surveyDBHandler.ts
```diff
@@ -52,6 +52,9 @@ export default class SurveyDBHandler {
       id: surveyContainerName,
       partitionKey: {
         paths: ['/callId']
+      },
+      uniqueKeyPolicy: {
+        uniqueKeys: [{ paths: ['/callId'] }]
```
rvenugopal-msft, line 57: should the unique key be based on callId and acsUserId?

rvenugopal-msft: You may need to handle the uniqueKey Constraint exception within your code, in case the duplicate exists. You may be handle the exception and return response as success 200. Some thoughts there on how to handle exception elegantly.

hansung-msft: If there is a exception throw from `upsert`, the express throw status code 500. Middleware already take care of it.

hansung-msft: I can put `acsUserId` as unique as well but it will not check `if callId AND acsUserId exist` since it only checks individual's uniqueness

rvenugopal-msft: The 500 is internal server error which is not clear and too generic for the client side to act upon. Since in this case the server knows the exception is from db and is from unique constraint, it should handle it elegantly and return a status code as may be 409(https://restfulapi.net/http-status-codes/) or just return 200 (as entity is already processed)

hansung-msft: Since this error will be from `upsert` procedure, Express middleware handles that it returns 500 generic error with message. Let me see what I can do for this.

rvenugopal-msft: I think that's may be fine within the context that we will anyways show "Thanks for submitting the survey" , so should not matter as we don't need to address for the unique key constraint violation on UI to keep it simple. So I am fine with default 500 error code there as of now

ralphgabrinao (dismissed):

server/src/databaseHandlers/surveyDBHandler.ts
hansung-msft: Fortunately, I was able to use the route to handle the error more gracefully. I just pushed new changes with test.

server/src/app.ts
```diff
@@ -68,6 +68,10 @@ app.use((err, req, res, next) => {
     return next(err);
   }
 
+  if (req.route.path === '/api/surveyResults') {
```
ralphgabrinao, line 71: I don't feel like it is necessary to handle errors for specific endpoints differently than others. For example, we don't do anything special for the `/api/token` and `/api/config` endpoints, so I am worried we are adding this unnecessarily and establishing a pattern that I think is not very scalable for future endpoints we might add.

I am completely fine without this part, and let Contoso deal with errors if they don't like the default 500 status code.

What are your thoughts on this? @hansung-msft @rvenugopal-msft

hansung-msft: Yea as we discussed, I am okay with generic error. If we all agree with generic error, I will revert the change. 

cc @rvenugopal-msft

rvenugopal-msft: Thats a fair point. I think that makes sense to me . If we want to do that we handle it as a generic controller logic not just for one endpoint.

hansung-msft: Sounds good! I will revert the changes and push it again. Thanks for sharing the opinions guys.

ralphgabrinao (approved):

dailam-msft (approved):
Looks good..

PR #120, opened 2022-11-01
Added routes for Rooms

hansung-msft:
# What

- This is the draft PR that does not have the tests yet.
# Why

# How Tested

I used Postman to send API to check if it returns desired response.

- Send POST request to `localhost:8080/api/testAppointments` and response should be looks like this:
  
  ```json
  {
      "roomId": "<some-rooms-id>",
      "participants": [
          {
              "id": <some-user-id>,
              "role": "Presenter"
          },
          {
              "id": <some-user-id>,
              "role": "Attendee"
          }
      ]
  }
  ```

- Then use above information to send GET request as below with query string:
  
  `localhost:8080/api/testAppointments/join?roomId=<rooms-id>&userId=<user-id>`

  Then you should have response as below:

  ```json
    {
        "participant": {
            "id": <user-id>,
            "role": "Presenter"
        },
        "token": <token-string>
    }
  ```

## Unit tests
image

# Process & policy checklist


server/src/controllers/roomsController.ts
```diff
@@ -0,0 +1,103 @@
...
+    const participantsList = await roomsClient.getParticipants(roomId as string);
+
+    // Check if the user is part of participants
+    const foundUserParticipant: RoomParticipant | undefined = participantsList.find(
+      (participant: RoomParticipant) => (participant.id as CommunicationUserIdentifier).communicationUserId === userId
+    );
+
+    if (foundUserParticipant) {
+      // Create token
+      const scopes: TokenScope[] = ['voip'];
```
allenplusplus, line 80: Do we need chat scope here as well?

server/src/controllers/roomsController.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    // Retrieve participants list
+    const participantsList = await roomsClient.getParticipants(roomId as string);
+
+    // Check if the user is part of participants
+    const foundUserParticipant: RoomParticipant | undefined = participantsList.find(
+      (participant: RoomParticipant) => (participant.id as CommunicationUserIdentifier).communicationUserId === userId
+    );
+
+    if (foundUserParticipant) {
```
allenplusplus, line 78: Suggestion to inverse the if check so we exit early. It's just for more readable code but it's not really important. Up to you if you want to change it. For example:

```
if (!foundUserParticipant)
{
    return 404;
}

// Retrieve token
```

This way the larger block of code for retrieving token doesn't need to be indented another level as it is outside the if block

server/src/routes/roomsRoutes.ts
```diff
@@ -0,0 +1,17 @@
...
+import { CommunicationIdentityClient } from '@azure/communication-identity';
+import { RoomsClient } from '@azure/communication-rooms';
+import { createRoom, joinRoom } from '../controllers/roomsController';
+
+export const roomsRouter = (identityClient: CommunicationIdentityClient, roomsClient: RoomsClient): express.Router => {
+  // Initialize router
+  const router = express.Router();
+
+  router.post('/', createRoom(identityClient, roomsClient));
+  router.get('/join', joinRoom(identityClient, roomsClient));
```
allenplusplus, line 14: I'm not sure if we should call this 'join'. To me the word join sounds like the participant is being added to the room. But this API is just getting token. Maybe we should call it 'token'? Also I would suggest we use POST because there is a resource (token) being created

server/src/controllers/roomsController.ts
```diff
@@ -0,0 +1,103 @@
...
+      })
+    );
+
+    // Formulate response
+    const response: CreateTestAppointmentResponse = {
+      roomId: createdRoom.id,
+      participants: participants
+    };
+
+    return res.send(response);
```
allenplusplus, line 56: Does this return 200? Would suggest 201 for created to mirror the rooms create API.

allenplusplus (dismissed):

server/src/controllers/roomsController.ts
hansung-msft: I am afraid that this is necessary.

hansung-msft: I will double check it tomorrow!

server/src/controllers/roomsController.ts
```diff
@@ -0,0 +1,103 @@
...
+          id: (foundUserParticipant.id as CommunicationUserIdentifier).communicationUserId,
+          role: foundUserParticipant.role as RoomParticipantRole
+        },
+        token: tokenResponse.token
+      };
+
+      return res.send(response);
+    }
+
+    return res.status(400).send(ERROR_NO_USER_FOUND_IN_ROOM);
```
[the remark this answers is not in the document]
hansung-msft: I was specifically choose 400. But according to this [MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404), I think it make sense to use 404. I will update it.

server/src/routes/roomsRoutes.ts
hansung-msft: That is a good point. I actually had similar question but different reason. I will speak with @ralphgabrinao tomorrow and leave the conclusion here.

server/src/controllers/roomsController.ts
hansung-msft: I believe the default returns 200. I also think 201 seems more suitable in here. I will update it tomorrow.

server/src/controllers/roomsController.ts
```diff
@@ -0,0 +1,103 @@
...
+export const joinRoom = (identityClient: CommunicationIdentityClient, roomsClient: RoomsClient) => async (
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction
+): Promise<any> => {
+  try {
+    const { roomId, userId } = req.query;
+
+    // Retrieve participants list
+    const participantsList = await roomsClient.getParticipants(roomId as string);
```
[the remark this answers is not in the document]
hansung-msft: Based on the screenshot below, it doesn't look like the `participantsList` can be null. It could be empty tho.

image

server/src/routes/roomsRoutes.ts
ralphgabrinao: Agreed, I like `/api/rooms/token`

server/src/controllers/roomsController.ts
hansung-msft: It looks like the `TokenScope` is required parameter for `getToken` as below:
image

server/src/routes/roomsRoutes.ts
hansung-msft: @ralphgabrinao -- Since we change this from `GET` to `POST` method, should we send the `userId` and `roomId` as body payload instead of query string?

ralphgabrinao: I think it will have to stay as query string because the idea is that users can join the room using a link: `/visit?userId=<myUserId>&roomId=<myRoomId>` 

If the userId and roomId are put in the body instead, it wouldn't be possible to join using a link like above.

ralphgabrinao: Yeah, putting in the body payload works. Then the e2e flow would be:

- User pastes this url in the browser: `myapp.com/visit?userId=<myUserId>&roomId=<myRoomId>`
- MeetingExperience component extracts the userId and roomId from query string
- MeetingExperience calls this endpoint and puts userId and roomId in the body payload to get token
- MeetingExperience renders CallComposite

server/src/controllers/roomsController.ts
allenplusplus: I mean like this `const scopes: TokenScope[] = ['voip', 'chat'];`

ralphgabrinao (dismissed):

server/src/constants.ts
```diff
@@ -22,6 +22,8 @@ export const VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME =
 export const VV_COSMOS_DB_CONNECTION_STRING = 'VV_COSMOS_DB_CONNECTION_STRING';
 export const VV_COSMOS_DB_NAME = 'VV_COSMOS_DB_NAME';
 
+export const ERROR_NO_USER_FOUND_IN_ROOM = 'The user is not participant of the room';
```
ralphgabrinao, line 25: nit: `The user is not a participant of the room`

server/src/routes/roomsRoutes.ts
```diff
@@ -0,0 +1,17 @@
...
+import { CommunicationIdentityClient } from '@azure/communication-identity';
+import { RoomsClient } from '@azure/communication-rooms';
+import { createRoom, joinRoom } from '../controllers/roomsController';
+
+export const roomsRouter = (identityClient: CommunicationIdentityClient, roomsClient: RoomsClient): express.Router => {
+  // Initialize router
+  const router = express.Router();
+
+  router.post('/', createRoom(identityClient, roomsClient));
+  router.post('/token', joinRoom(identityClient, roomsClient));
```
allenplusplus, line 14: Minor one: rename joinRoom to createToken or something like that

allenplusplus (approved):

ralphgabrinao (approved):

server/src/controllers/roomsController.ts
hansung-msft: I don't think Rooms supports chat. Also, I don't see the `chat` option in the sample and [documentation](https://learn.microsoft.com/en-us/azure/communication-services/quickstarts/access-tokens?tabs=windows&pivots=programming-language-javascript#issue-access-tokens).

image

server/src/routes/roomsRoutes.ts
hansung-msft: I will update it to `getToken` instead of `joinRoom`

ralphgabrinao (dismissed):

e2e-tests/tests/common/utils.ts
```diff
@@ -15,7 +15,7 @@ export const buildUrl = (
   }
 };
 
-export const DELAY_MS = 1200;
+export const DELAY_MS = 1500;
```
tejaswip-msft, line 18: this is not needed now as we removed /book snapshot testing

ralphgabrinao (approved):

allenplusplus (approved):

PR #121, opened 2022-11-01
Show spinner while inserting survey response

dailam-msft:
# What
1) Added spinner to the submit/continue button on PostCallOneQuestionPoll component. Adjusted the style of Submit button accordingly
The button will be disabled and the spinner will be shown while insertion is done
Once the db call is completed the window will be redirected to /book page.

2) corrected a test name and added a conole.log error in case the db insertion has issues.

# Why
No transitional screen/state was present after submitting the survey response

# How Tested
Locally loaded client and server 
spinner on continue

# Process & policy checklist


client/src/components/postcall/PostCallOneQuestionPoll.tsx
```diff
@@ -20,10 +20,11 @@ export const PostCallOneQuestionPoll: React.FunctionComponent<PostCallOneQuestio
   props: PostCallOneQuestionPollProps
 ) => {
   const [pollResponse, setPollResponse] = useState<boolean | string | number>();
+  const [isSubmittingResponse, setIsSubmittingResponse] = useState<boolean>(false);
 
   const submitSurveyResponse = async (): Promise<void> => {
     const { acsUserId, callId, meetingLink } = props;
-
+    setIsSubmittingResponse(true);
     await submitSurveyResponseUtil(acsUserId, pollResponse, meetingLink, callId);
```
allenplusplus, line 28: We might want to add a try catch here to setIsSubmittingResponse(false) if the submitSurveyResponseUtil() call throws an exception. Otherwise if exception is thrown, UI will be stuck spinning and user cannot try re-submitting. For example:

```
setIsSubmittingResponse(true);
try
{
    await submitSurveyResponseUtil(acsUserId, pollResponse, meetingLink, callId);
    window.location.replace('/book');
}
catch (e)
{
    setIsSubmittingResponse(false);
}
```

client/src/utils/PostCallUtil.ts
```diff
@@ -22,6 +22,7 @@ export const submitSurveyResponseUtil = async (
       })
     });
     if (res.status !== 200) {
+      console.log('Error during insertion');
```
allenplusplus, line 25: Is this logging for testing and should be removed?

allenplusplus (approved):

client/src/utils/PostCallUtil.ts
dailam-msft: I added this logging function just so Contoso would get some indication that something went wrong. My thought was to have a minimized version of Error Handling of sorts. Do you think this would be bad practice? I can remove it if so.

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: I see your point, but I do have a try...catch within the **submitSurveyResponseUtil** function. Also since we have window redirect I feel that this code may not really execute.

client/src/styles/Survey.styles.ts
```diff
@@ -63,7 +63,6 @@ export const pollTitleStyle = {
 };
 
 export const surveySubmitButtonStyles = {
-  width: '20rem',
```
ralphgabrinao, line 66: Curious why this is being removed?

dailam-msft: When the button was that wide I had the spinner come up on the extreme left while the text was aligned in center. It was visually jarring and this is better. Mentioned this change in button size to UX on the channel too.

client/src/components/postcall/PostCallOneQuestionPoll.tsx
allenplusplus: I like it here because it allows UI to not have to depend on the internal implementation of submitSurveyResponseUtil and UI is more resilient to future changes. But up to you if you want to change it or leave it this way.

client/src/utils/PostCallUtil.ts
allenplusplus: I don't think it is bad to log to console. Only that it might get noisy for the developer. Ideally we would have a logging system where developers can configure the level of logging and have a way to reduce the noise. But we don't have that right now and also it might look like we forgot to remove a debug console.log.

client/src/styles/Survey.styles.ts
ralphgabrinao: This piece of code works: 

2022-11-02 14_12_51-PostCallOneQuestionPoll tsx - communication-services-virtual-visits-js-dailasm-s

2022-11-02 14_12_11-Book and 9 more pages - Work - Microsoft​ Edge

This way the button width matches the text box width and also retains consistency with the Builder preview.

dailam-msft: Thanks! Have made this change

client/src/components/postcall/PostCallOneQuestionPoll.tsx
dailam-msft: hmm Ok, have made this change! thanks!

client/src/utils/PostCallUtil.ts
dailam-msft: ok, removed the log and just added a comment instead

allenplusplus (approved):

ralphgabrinao (approved):

PR #125, opened 2022-11-03
Default to postcall undefined if unable to initialize CosmosDb

ralphgabrinao:
# What
Gracefully handle in case CosmosDb cannot be initialized so that the app continues to run.

Remove redundant tests for createSurveyDBHandler.

# Why
If Cosmos cannot be initialized, the app crashes. If such a case, display the default post call experience from UI Library instead.

# How Tested
- Change the Cosmos Db connection string to something invalid and confirm that the /api/config endpoint returns the config without postcall
- Change the Cosmos Db connection string to a valid one and confirm that the /api/config endpoint returns the config with postcall

# Process & policy checklist


hansung-msft:
I can see this silently handles on the backend well. What do you think about we give them some kind of warning?

hansung-msft (approved):

dailam-msft (approved):

PR #140, opened 2022-11-24
Dailasm/uifixes homepage

dailam-msft:
# What

Fixed following Ui Issues . Also fixed snapshots for Home and Book (as FAQ section is same for both pages)
	

1. Font for link - verified that segoe is fine - done 	
2. Learn more text - fixed to show "…" 	
3. Font size 1rem - done 	
4. Icons - blue - done 	
5. 2rem margin top above container with all text after calendar illustration - done 	
6. Flex width sizing for the buttons for while we have two they use all the container width and when we add the third one they resize - done 	
7. each of the links under "Learn more..." open a new tab - done 	
8. the alignment of the join from link button as it is not centered - done 

# Why
UI Issues

# How Tested
locally and deployed - 
image

NO

NO


client/src/styles/Home.styles.ts
```diff
@@ -56,22 +55,42 @@ export const imageStyles = {
...
+export const joinLinkButtonStyles = {
   root: {
-    width: '11.8125rem',
     borderColor: '#E1DFDD',
-    fontSize: '.75rem'
+    fontSize: '1rem'
   }
 };
 
+export function calendarIconStyles(theme: PartialTheme | Theme | undefined): IIconProps {
```
hansung-msft, line 80: Does `Theme` ever be in `undefined`?

client/src/styles/Home.styles.ts
```diff
@@ -56,22 +55,42 @@ export const imageStyles = {
...
+  return {
+    iconName: 'Calendar',
+    styles: {
+      root: { color: theme?.palette?.themePrimary }
```
hansung-msft, line 84: I think once we verified that `Theme` can not be `undefined`, then we can remove `?`here.

client/src/styles/Home.styles.ts
```diff
@@ -115,18 +133,15 @@ export const layerHostStyles: CSSProperties = {
...
 
-export const joinLinkButtonStyles = {
-  root: {
-    paddingRight: '1.875rem',
-    width: '11.8125rem',
-    borderColor: '#E1DFDD',
-    fontSize: '.75rem'
-  }
-};
+export function linkIconStyles(theme: PartialTheme | Theme | undefined): IIconProps {
```
hansung-msft, line 136: Here too!

client/src/styles/Home.styles.ts
dailam-msft: nice catch 😄 removed undefined

client/src/styles/Home.styles.ts
dailam-msft: done!

client/src/styles/Home.styles.ts
dailam-msft: done.

hansung-msft:
It looks like we might need some space between the banner and `Hello` title.

This is from the Figma design:
image

hansung-msft:
> 

Nvm fort this!

hansung-msft (approved):

client/src/components/home/Home.tsx
```diff
@@ -1,27 +1,29 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { DefaultButton, PartialTheme, StackItem, Theme } from '@fluentui/react';
+import { DefaultButton, ImageFit, PartialTheme, StackItem, Theme } from '@fluentui/react';
 import { Stack, Text, Image } from '@fluentui/react';
```
tejaswip-msft, line 5: Lets rename this file, there is another Home.tsx and it might get confusing, we can rename this to HomeComponent.tsx

client/src/styles/Home.styles.ts
```diff
@@ -12,7 +12,7 @@ export const fullScreenStyles = {
   }
 };
```
tejaswip-msft, line 13: do we want the page to look like visit , with the gap below
image

Currently the home page has the white container extending till end 
image

Also the theme is not reflected in the header for /home, i changed to red, as u can see in /visit the header is red

dailam-msft: Home page extending till the end of page is as per the figma, and personally i think this looks better, looks more anchoring and larger and like a true "home" page.. The theme issue am looking into it

dailam-msft: Thank you, this was a miss in providing config.theme to the ThemeProvide. Have pushed a fix!

client/src/components/home/Home.tsx
dailam-msft: have changed this and updated references to it and pushed a fix!

client/src/styles/Home.styles.ts
hansung-msft: This is good find. I didn't change the color and tested it...

tejaswip-msft (approved):

client/src/styles/Home.styles.ts
dailam-msft: image
Yes indeed :) . Confirmed with UX that this behaviour is desirable/ expected

hansung-msft (approved):

PR #146, opened 2022-12-02
Start call - presenter and bugfixes

dailam-msft:
# What
1)Start call button creates a room with presenter in it
2)addressed e2e comment to replace url with constant
3) fixed bug for navigation using browser buttons

# How Tested
unit tested , Locally and deployed - 

image

### DIFF
--- a/client/src/components/home/HomeComponent.tsx
+++ b/client/src/components/home/HomeComponent.tsx
@@ -1,30 +1,49 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { DefaultButton, ImageFit, PartialTheme, StackItem, Theme } from '@fluentui/react';
+import { DefaultButton, IContextualMenuProps, ImageFit, PartialTheme, StackItem, Theme } from '@fluentui/react';
 import { Stack, Text, Image } from '@fluentui/react';
 import imageCalendar from '../../assets/lightCalendarSymbol.png';
 import {
   btnStackStyles,
-  bookAppointmentButtonStyle,
   calendarIconStyles,
   containerMarginTop2rem,
   containerStyles,
   font16pxStyle,
   fullScreenStyles,
   innerContainer,
-  joinLinkButtonStyles,
   lineHeight22px,
   lineHeight28px,
-  linkIconStyles
+  linkIconStyles,
+  videoIconStyles,
+  buttonStyles
 } from '../../styles/Home.styles';
 import { FrequentlyAskedQuestions } from '../FrequentlyAskedQuestions';
 import { LearnMoreItem } from '../LearnMoreItem';
+import { createRoom } from '../../utils/CreateRoom';
+import { RoomParticipantRole } from '../../models/RoomModel';
 
 export interface HomeProps {
   companyName: string;
   theme: PartialTheme | Theme;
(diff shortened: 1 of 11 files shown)


client/src/components/home/HomeComponent.tsx
```diff
@@ -42,21 +61,31 @@ export const HomeComponent = (props: HomeProps): JSX.Element => {
...
+                />
+              </StackItem>
+              <StackItem>
+                <DefaultButton
+                  text="Start a call"
+                  splitButtonAriaLabel="See 2 options"
+                  styles={buttonStyles}
+                  iconProps={videoIconStyles(props.theme)}
+                  aria-roledescription="split button"
+                  menuProps={menuProps}
```
ralphgabrinao, line 80: This should just be a regular button based on Jessica's guidance.

client/src/utils/FetchRoomsResponse.ts
```diff
@@ -25,6 +25,25 @@ export const fetchRoomsResponse = async (roomId: string, userId: string): Promis
 
   const responseContent = await response.text();
   const roomResponse = JSON.parse(responseContent);
-
   return roomResponse;
 };
+
+export async function fetchRoom(): Promise<CreateRoomResponse> {
```
ralphgabrinao, line 31: Arrow syntax here. Also let's make the function name more accurate about what it is doing `createRoom`

client/src/utils/CreateRoom.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+import { RoomParticipant, RoomParticipantRole } from '../models/RoomModel';
+import { fetchRoom } from './FetchRoomsResponse';
+
+export async function createRoom(userRole: string): Promise<string> {
```
ralphgabrinao, line 6: Arrow syntax here

client/src/components/home/HomeComponent.tsx
```diff
@@ -1,31 +1,50 @@
...
+      text: 'as host (presenter)',
+      onClick: (): void => {
+        callCreateRoom(RoomParticipantRole.presenter);
+      }
+    }
+  ]
+};
+
+const callCreateRoom = async (role: RoomParticipantRole): Promise<void> => {
+  const redirectUrl = await createRoom(role);
```
ralphgabrinao, line 44: Let's also add error handling in case the API call fails

client/src/utils/CreateRoom.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    createRoomSpy.mockImplementation(
+      async (): Promise<CreateRoomResponse> => {
+        throw new Error('test error');
+      }
+    );
+
+    const mockStatus = 400;
+    const mockUserRole = 'userRole';
+
+    (fetch as jest.Mock).mockImplementation(() => {
```
ralphgabrinao, line 58: My hunch is because the `createRoom` function is being mocked, the test doesn't even go inside that function. Therefore I think mocking fetch is not necessary and can be removed.

tejaswip-msft (approved):
LGTM

ralphgabrinao (approved):

PR #159, opened 2022-12-15
Add instructions for copying invite link for Attendee in a Rooms call

ralphgabrinao:
# What
Implement MessageBar for Copy Invite Link instructions, only visible to the Presenter.

Also fixing an overflow issue in GenericContainer so that the height adjusts based on the size of the contents.

If formFactor is mobile, do not show the MessageBar since there is little real estate.

# Why
As a Presenter in a Rooms call, I should see instructions on how to copy the invite link for the Attendee.

# How Tested
Join the Room as the Presenter and confirm that the MessageBar is visible. Join the Room as the Attendee and confirm that the MessageBar is not visible. The MessageBar should also be invisible if the formFactor is mobile.

MessageBar:
2022-12-14 15_26_07-MeetingExperienceUtil ts - communication-services-virtual-visits-js - Visual Stu

GenericContainer overflow:
2022-12-14 17_11_31-RoomsMeetingExperience tsx - communication-services-virtual-visits-js - Visual S

# Process & policy checklist


client/src/utils/MeetingExperienceUtil.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { CallCompositePage } from '@azure/communication-react';
```
ralphgabrinao, line 4: https://github.com/Azure/communication-ui-library/blob/b3eb013be390e5dd3895fccb995b1950b1a93693/packages/react-composites/src/composites/CallComposite/CallComposite.tsx#L217

client/src/components/rooms/RoomsMeetingExperience.tsx
```diff
@@ -21,32 +23,51 @@ export interface RoomsMeetingExperienceProps {
...
-    const _createAdapters = async (): Promise<void> => {
+    const createAdapters = async (): Promise<void> => {
       try {
         const adapter = await _createCustomAdapter({ communicationUserId: userId }, credential, displayName, locator);
-        if (userRole !== RoomParticipantRole.presenter && postCall?.survey.type) {
-          adapter.on('callEnded', () => {
-            setRenderPostCall(true);
-          });
+
+        const postCallEnabled = isRoomsPostCallEnabled(userRole, postCall);
```
tejaswip-msft, line 47: nit: if we are using this only at one place, do we want to put this in a const

tejaswip-msft (approved):

dailam-msft (approved):

PR #162, opened 2022-12-20
Join a room from "Join Call" screen

dailam-msft:
# What
Changes to be able to use "Join Call" screen to join a valid room

# Why
participants should be able to join a call directly as well as through "Join Call" screen

# How Tested
unit tested, locally and deployed -
From the deployed app - 
1) go to /home, then click "Start as Presenter" . 
2) Join the room as presenter and click on the people icon button and copy the invite link . 
3) Launch a new window with same deployed link at /home and click "Join from Link" button. 
4) Paste the copied link into the textbox and user will be added into same room as presenter (from step 1 above)


client/src/utils/GetMeetingLink.ts
```diff
@@ -75,6 +75,17 @@ export const isValidTeamsLink = (teamsMeetingLink: string): boolean => {
   return true;
 };
 
+export const isValidRoomsLink = (roomsMeetingLink: string): boolean => {
+  if (
+    !roomsMeetingLink.startsWith(window.location.origin + '/visit') ||
+    !roomsMeetingLink.includes('roomId') ||
+    !roomsMeetingLink.includes('userId')
```
ralphgabrinao, line 82: We can be more precise here by using a regex.

For instance, just using `.includes()` here would still render this a valid rooms link: `https://localhost:3000/visit_roomId-%^&userId`

client/src/utils/GetMeetingLink.test.ts
```diff
@@ -135,4 +136,19 @@ describe('getMeetingLink', () => {
     const result = makeRoomsJoinUrl('mockRoomId', 'mockUserId');
     expect(result).toBe('/visit?roomId=mockRoomId&userId=mockUserId');
   });
+
+  test('should return true if RoomsMeetingUrl is valid', () => {
```
tejaswip-msft, line 140: lets add another test case to cover the "false" case as well

client/src/components/JoinMeeting.test.tsx
```diff
@@ -4,18 +4,18 @@
 import { PrimaryButton, TextField } from '@fluentui/react';
 import { mount } from 'enzyme';
 import { generateTheme } from '../utils/ThemeGenerator';
-import { JoinTeamsMeeting } from './JoinTeamsMeeting';
+import { JoinMeeting } from './JoinMeeting';
```
tejaswip-msft, line 7: we can do a for each for all the test cases in this file and add for rooms meeting link as well

client/src/Visit.tsx
```diff
@@ -57,11 +57,13 @@ export const Visit = (): JSX.Element => {
     const appendMeetingLinkToUrl = (): void => {
       window.history.pushState({}, document.title, window.location.href + link);
     };
-
-    appendMeetingLinkToUrl();
-
-    const meetingLinkLocator = _getTeamsMeetingLinkLocator(link);
-    setTeamsMeetingLinkLocator(meetingLinkLocator);
+    const boolValidRoomsLink = isValidRoomsLink(link);
```
tejaswip-msft, line 60: umm, name better? or since used just once we can just move it directly to the if 
if (isValidRoomsLink(link)) {
}

client/src/utils/GetMeetingLink.test.ts
ralphgabrinao: i.e. if the link:
- does not start with window.location.origin + '/visit'
- has no roomId query parameter
- has no userId query parameter
- query parameters are invalid format: `http://localhost:3000/visit?123=roomId&userId=456` and `http://localhost:3000/visit?roomId=123&456=userId`

dailam-msft: Test coverage currently being 100% I think we should skip this - 
image

client/src/Visit.tsx
dailam-msft: done!

client/src/utils/GetMeetingLink.ts
dailam-msft: done!

client/src/components/JoinMeeting.test.tsx
dailam-msft: Let's discuss! I don't think I completely understand what you're suggesting 😃

client/src/utils/GetMeetingLink.test.ts
ralphgabrinao: I am surprised that it is showing 100% when it is clear that we are not covering that path. I think it is because the success path still passes through the entire `if` block.

After organizing the tests by putting each unit in its own `describe` block, we can see that this file is actually not 100% covered:
2022-12-20 13_29_40-GetMeetingLink test ts - communication-services-virtual-visits-js - Visual Studi

Please feel free to refer to/copy-paste the changes here:
- https://github.com/Azure-Samples/communication-services-virtual-visits-js/commit/d4381a56f985e885d0e28e6283ff5bf740a4f857

This adds the m
2022-12-20 13_48_03-GetMeetingLink test ts - communication-services-virtual-visits-js - Visual Studi
issing test cases for this file, and organizes each function in its own describe block.

dailam-msft: Oh nice! great find.. Can you push your changes to this same branch.. We can merge it out quicker..

ralphgabrinao: Sure thing. I hope it would still allow me to approve this though. We shall see...

ralphgabrinao (approved):

ralphgabrinao (approved):

tejaswip-msft (approved):

client/src/components/JoinMeeting.test.tsx
dailam-msft: Done!

PR #208, opened 2025-04-02
[Transcription & Summarization] Add new routes for server to enable transcription and summarization

dmceachernmsft:
# What

Adds new routes to support Transcription and meeting summarization
- startTranscription
- stopTranscription
- connectToRoomsCall
- fetchTranscription
- startCallWithTranscription
- callAutomationEvent

Also adds a websocket server to the server application to handle callAutomation events

# Why

Allows the server to handle the different events from callAutomation and provide Transcription to Rooms users

# How Tested
Built and tested locally More PR's to come with client functionality and testing

# Process & policy checklist

- [ x ] I have updated the project documentation to reflect my changes if necessary.


server/src/app.ts
```diff
@@ -43,6 +55,48 @@ app.get('/visit', (_, res) => {
...
+ */
+app.use('/api/callAutomationEvent', cors(), callAutomationEvent);
+
+/**
+ * route: /summarizeTranscript
+ * purpose: Sends transcript to AI summarization service
+ */
+app.use('/api/summarizeTranscript', cors(), summarizeTranscript);
+
+app.use('/api/events', cors(), events);
```
JamesBurnside, line 98: add comment to follow pattern of other routes

JamesBurnside, line 98: `/events` is very generic, there is also `/callAutomationEvent`, rename this to be more specific on its purpose

server/src/routes/callAutomationEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+import { CALLCONNECTION_ID_TO_CORRELATION_ID } from '../utils/callAutomationUtils';
+
+const router = express.Router();
+
+router.post('/', async function (req, res) {
+  try {
+    if (req.body[0].type === 'Microsoft.Communication.CallConnected') {
+      console.log('/automationEvent received', req.body);
+      /**
+       * if the call already exists in the mapping we don't want to start a
```
JamesBurnside, line 14: dangling comment

server/src/routes/callAutomationEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+       */
+      if (CALLCONNECTION_ID_TO_CORRELATION_ID[req.body[0].data.callConnectionId]) {
+        console.log('CallConnectionId already exists in mapping');
+
+        res.status(200).end();
+        return;
+      }
+      /**
+       * We want to make a mapping here between the callConnectionId and the correlationId
```
JamesBurnside, line 23: nit: comments should use imperative tense
```suggestion
        * Make a mapping here between the callConnectionId and the correlationId
```

server/src/routes/callAutomationEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT license.
+
+import * as express from 'express';
+import { CALLCONNECTION_ID_TO_CORRELATION_ID } from '../utils/callAutomationUtils';
+
+const router = express.Router();
+
+router.post('/', async function (req, res) {
+  try {
+    if (req.body[0].type === 'Microsoft.Communication.CallConnected') {
```
JamesBurnside, line 11: Avoid undefined reference exception
```suggestion
    if (req.body[0]?.type === 'Microsoft.Communication.CallConnected') {
```

server/src/routes/fetchTranscript.ts
```diff
@@ -0,0 +1,41 @@
...
+
+router.post('/', async function (req, res, next) {
+  const { serverCallId }: FetchTranscriptRequest = req.body;
+  console.log('Fetching transcript for call:', serverCallId, 'available calls:', Object.keys(TRANSCRIPTION_STORE));
+  /**
+   * callId here is the correlationId in the Automation event saying transcription has started
+   * we need to use this to get the call connectionId from the callAutomation client
+   */
+  const connectionId = Object.keys(CALLCONNECTION_ID_TO_CORRELATION_ID).find((key) =>
+    CALLCONNECTION_ID_TO_CORRELATION_ID[key].serverCallId.includes(serverCallId)
```
JamesBurnside, line 24: should this be `includes` or `===`?

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+ * because all of the clients that join can have different callId's and we need to make sure we are
+ * pulling the correct transcription data.
+ */
+export const TRANSCRIPTION_STORE: { [key: string]: Partial<CallTranscription> } = {};
+/**
+ * Used to map between the call connection id and the correlation id from both transcription and
+ * call automation events.
+ */
+export const CALLCONNECTION_ID_TO_CORRELATION_ID: {
+  [key: string]: { correlationId?: string; serverCallId: string };
```
JamesBurnside, line 147: consider using a `Map` type instead of this raw map (the reason we have used a raw map in the stateful client is to make it serializable in state but we dont need that here). It will help make the operations performed on this easier to read

```ts
export const CallConnectionToCorrelationIdMap = new Map<string, { correlationId?: string; serverCallId: string }>
```

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+ * We should re-write all the fetches based on the callId to use the serverCallId. This is
+ * because all of the clients that join can have different callId's and we need to make sure we are
+ * pulling the correct transcription data.
+ */
+export const TRANSCRIPTION_STORE: { [key: string]: Partial<CallTranscription> } = {};
+/**
+ * Used to map between the call connection id and the correlation id from both transcription and
+ * call automation events.
+ */
+export const CALLCONNECTION_ID_TO_CORRELATION_ID: {
```
JamesBurnside, line 146: There's a lot of mutable global variables in this file. Consider writing a Manager class that encapsulates a lot of these raw entries, and then could also make them private but expose accessor functions, perform any lifecycle management etc.
(Can consider as a future improvement, don't need to block this PR on it).

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+  console.log('Starting transcription for call:', callConnectionId);
+  console.log(CALLCONNECTION_ID_TO_CORRELATION_ID);
+  const callConnection = await getCallAutomationClient().getCallConnection(callConnectionId);
+
+  return await callConnection.getCallMedia().startTranscription(options);
+};
+
+export const stopTranscriptionForCall = async (callConnectionId: string): Promise<void> => {
+  console.log('Stopping transcription for call:', callConnectionId);
+  console.log(CALLCONNECTION_ID_TO_CORRELATION_ID);
```
JamesBurnside, line 130: Check if all the console logs are still necessary, if so, combine the double ones into a single console log

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+ * This object is keyed off the callId and contains the communicationUserId and displayName of the remote participants
+ *
+ * Keeps track of all participants who have ever joined the call so we can show their display name in the transcription and summary.
+ */
+export const REMOTE_PARTICIPANTS_IN_CALL: {
+  [key: string]: { communicationUserId: string; displayName: string }[];
+} = {};
+
+/**
+ * used to store the local participant in the call
```
JamesBurnside, line 161: Does this only support a single local participant? What if two different calls are happen simultaneously?

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+    CALLCONNECTION_ID_TO_CORRELATION_ID[key].serverCallId.includes(serverCallId)
+  );
+  if (!connectionId) {
+    return undefined;
+  }
+  const correlationId = CALLCONNECTION_ID_TO_CORRELATION_ID[connectionId]?.correlationId;
+  if (!correlationId) {
+    return undefined;
+  }
+  return TRANSCRIPTION_STORE[correlationId] as CallTranscription;
```
JamesBurnside, line 181: How is this `CallTranscription` and not `Partial<CallTranscription>` now?

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+export const checkIfTranscriptionStarted = (serverCallId: string): boolean => {
+  console.log('Checking if transcription started for call:', serverCallId);
+  const connectionId = Object.keys(CALLCONNECTION_ID_TO_CORRELATION_ID).find((key) =>
+    CALLCONNECTION_ID_TO_CORRELATION_ID[key].serverCallId.includes(serverCallId)
+  );
+  if (!connectionId) {
+    return false;
+  }
+  const correlationId = CALLCONNECTION_ID_TO_CORRELATION_ID[connectionId]?.correlationId;
+  return correlationId ? !!TRANSCRIPTION_STORE[correlationId] : false;
```
JamesBurnside, line 196: something that a manager class would help clean up, instead of
```
  const correlationId = CALLCONNECTION_ID_TO_CORRELATION_ID[connectionId]?.correlationId;
  return correlationId ? !!TRANSCRIPTION_STORE[correlationId] : false;
```
we could have
```
return !!TranscriptionManager.getTranscriptions(correlationId);
```

server/src/utils/summarizationUtils.ts
```diff
@@ -0,0 +1,281 @@
...
+   * - analysisInput: An object that contains the conversation items.
+   * - tasks: An array of task objects that define the tasks to be performed on the conversation items.
+   * Each task object contains the following properties:
+   *   - taskName: A string that represents the name of the task.
+   *   - kind: A string that represents the type of task.
+   *   - parameters: An object that contains the parameters for the task.
+   *     - summaryAspects: An array of strings that represent the aspects of the summary to be generated.
+   * The response will contain the operation location URL, which can be used to poll for the job status.
+   */
+  console.log('localeCode:', localeCode);
```
JamesBurnside, line 130: remove console log

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+  const res = await getCallAutomationClient().connectCall(
+    {
+      kind: 'serverCallLocator',
+      id: serverCallId
+    },
+    callAutomationConfig?.ServerHttpUrl ?? '',
+    options
+  );
+  console.log('Connect call result', res);
+  const callConnection = res.callConnection;
```
JamesBurnside, line 113: remove unused variable

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
+// Copyright (c) Microsoft Corporation.
```
JamesBurnside, line 1: Suggestion. Would be great to have in the readme how this works if a developer is going to make an update to it.
There's async events from the websockets, various STORE_MAPS and its really hard to piece everything together reading the code as its not a linear feature.
Some sort of flowchart diagram would be very helpful

server/src/utils/callAutomationUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+  const callbackUri = callAutomationConfig?.ServerHttpUrl;
+
+  /**
+   * Call automation needs to create the call.
+   */
+
+  const automationClient = getCallAutomationClient();
+  const roomsLocator: CallLocator = { kind: 'roomCallLocator', id: roomId };
+  const connectedCallResult = await automationClient.connectCall(roomsLocator, callbackUri ?? '', options);
+  console.log((await connectedCallResult.callConnection.getCallConnectionProperties()).serverCallId);
```
mgamis-msft, line 64: Was this console just for debugging?

mgamis-msft (approved):

server/src/routes/fetchTranscript.ts
dmceachernmsft: These comparisons should be includes as the callAutomation return for the serverCallId is slightly different than the one from the CallingSDK. They add random amounts of '=' characters to the end of the id in their events so we need to see if the serverCallid from the client is a substring. There is a bug right now from working with the automation engineers where the correlation id's coming from the transcription service are not always right depending on who started the call and started transcription. 

tldr: going with the serverCallId for now but would like to be using the correlationId in the future.

server/src/utils/callAutomationUtils.ts
dmceachernmsft: absolutely. This will come in a future PR closer to the end.

server/src/utils/callAutomationUtils.ts
dmceachernmsft: yes a lot of them are will make sure to clean up anything not needed but some are useful for developers to see when running the sample

server/src/utils/callAutomationUtils.ts
dmceachernmsft: this is a good point we should be processing all the participants together. this is a work in progress for adding the names to the downloaded transcript so I will remove this and update it so that we are just tracking all the participants together.

server/src/utils/callAutomationUtils.ts
dmceachernmsft: the map can just be a CallTranscript now. I think there was originally some issues when setting up the metadata what the value would be for the transcription data itself and by making sure both are set when we start a new call and get that first event we can get rid of the partial

server/src/routes/fetchTranscript.ts
JamesBurnside: > These comparisons should be includes as the callAutomation return for the serverCallId is slightly different than the one from the CallingSDK.

Are the teams aware of this to fix it?

server/src/routes/startTranscription.ts
```diff
@@ -0,0 +1,36 @@
...
+    res.status(500).send('Error starting transcription');
+    sendEventToClients('TranscriptionError', {
+      serverCallId
+    });
+    return;
+  }
+
+  res.status(200).end();
+
+  // Send SSE event to clients
```
dmceachernmsft, line 33: clean this up

server/src/utils/TranscriptionManager.ts
```diff
@@ -0,0 +1,201 @@
...
+    }
+    return this.transcriptionStore.get(correlationId);
+  }
+
+  /**
+   * Stores the transcription metadata for the call
+   * @param data - meta data to create the record for the transcription
+   * @returns
+   */
+  public storeTranscriptionMetaData(data: TranscriptionMetadata): void {
```
dmceachernmsft, line 75: oops

.github/workflows/e2e-tests.yml
```diff
@@ -62,6 +62,7 @@ jobs:
           Write-Host "Launching Cosmos DB Emulator"
           Import-Module "$env:ProgramFiles\Azure Cosmos DB Emulator\PSModules\Microsoft.Azure.CosmosDB.Emulator"
           Start-CosmosDbEmulator
+          Start-Sleep -Seconds 300
```
JamesBurnside, line 65: I get this unblocks us today but we _really_ should not have this

server/src/app.ts
```diff
@@ -66,6 +141,42 @@ app.get('/api/config', configController(config));
...
+      } else {
+        console.log(`Client ${index} connection ended, removing from clients list`);
+        const clientIndex = clients.indexOf(client);
+        if (clientIndex !== -1) {
+          clients.splice(clientIndex, 1);
+        }
+      }
+    } catch (error) {
+      console.error(`Failed to send event to client ${index}:`, error);
+      // Remove failed client from the original array
```
JamesBurnside, line 170: Is it 100% of any error we want to ignore that client? What if the client went into a tunnel for 2 mins, would it error then but reconnect?

JamesBurnside (dismissed):

server/src/app.ts
dmceachernmsft: so the client will initiate a reconnect when it has a internet connection again. this is so the server isn't blasting bits at a dead connection. EventSource will attempt to reconnect to the subscription on the client side as part of the API

JamesBurnside (approved):

PR #212, opened 2025-04-08
[Transcription] Add transcription and meeting summary

dmceachernmsft:
# What

Adds the transcription and meeting summary functionality to Sample
# Why

Allows for users to transcribe and summarize meetings

# How Tested
Built and tested locally
image
image

# Process & policy checklist

- [ x ] I have updated the project documentation to reflect my changes if necessary.
- [ x ] I have read the [CONTRIBUTING](CONTRIBUTING.md) documentation.

- [ x ] I have tested the deployed app end to end


client/src/components/rooms/transcriptionNotifications/CallSummaryTile.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  </Text>
+                </Stack>
+                {summary && <Text styles={{ root: { fontWeight: 400 } }}>{summary}</Text>}
+                <DefaultButton
+                  text="Transcript"
+                  onClick={async () => {
+                    if (summary && serverCallId) {
+                      const transcript = await fetchTranscript(serverCallId);
+                      if (!transcript) {
+                        console.error('Failed to fetch transcript');
```
edwardlee-msft, line 83: Is there is subcode/code errors that are surfaced from the fetchTranscript?

client/src/components/rooms/RoomsMeetingExperience.tsx
```diff
@@ -2,33 +2,228 @@
...
+  const theme = useTheme();
+  const callAutomationStarted = useRef(false);
+  const eventSourceRef = useRef<EventSource | null>(null);
+
+  useEffect(() => {
+    let eventSource: EventSource | null = null;
+
+    if (serverCallId && transcriptionFeatureEnabled) {
+      // Create EventSource connection when serverCallId is available. The URL provided here is for your server.
+      eventSource = new EventSource(`http://localhost:8080/api/notificationEvents`);
```
edwardlee-msft, line 84: The comment here doesn't feel distinct enough to warrant attention. 
Can we better indicate this may need to be changed to their URL?

client/src/components/rooms/RoomsMeetingExperience.tsx
```diff
@@ -2,33 +2,228 @@
...
+  }, [serverCallId]);
+
+  useEffect(() => {
+    if (!eventSourceRef.current && !callConnected && transcriptionFeatureEnabled) {
+      return;
+    }
+    eventSourceRef.current?.addEventListener('TranscriptionStarted', (event) => {
+      const parsedData = JSON.parse(event.data);
+      if (parsedData.serverCallId.includes(serverCallId)) {
+        console.log('Transcription started', event.data);
```
edwardlee-msft, line 123: I noticed we have a bunch of console logs, these definitely can be useful but are all of them necessary?
For functions and events that do return something, or have some kind of visible action, does it need console logs?

client/src/components/rooms/RoomsMeetingExperience.tsx
```diff
@@ -2,33 +2,228 @@
...
+  const [summary, setSummary] = useState<SummarizeResult>();
+  const [summarizationStatus, setSummarizationStatus] = useState<'None' | 'InProgress' | 'Complete'>('None');
+  const [statefulClient, setStatefulClient] = useState<StatefulCallClient>();
+  const [callAgent, setCallAgent] = useState<DeclarativeCallAgent>();
+  const [call, setCall] = useState<Call | TeamsCall | undefined>();
+  const [callAdapter, setCallAdapter] = useState<CommonCallAdapter>();
+  const [customNotications, setCustomNotifications] = useState<ActiveNotification[]>([]);
+  const [callConnected, setCallConnected] = useState(false);
+  const [serverCallId, setServerCallId] = useState<string | undefined>(undefined);
+  const [transcriptionStartedByYou, setTranscriptionStartedByYou] = useState(false);
```
edwardlee-msft, line 71: Wow that's a lot of states! hahah. 
Do we want to have a state manager, or at the bare minimum have these alphabetized ?

client/src/components/rooms/RoomsMeetingExperience.tsx
JamesBurnside: Agreed, does this break when deployed? We often use an environment variable for raw urls, or if its relative can it just be 
```suggestion
      eventSource = new EventSource(`/api/notificationEvents`);
```
^would something like that work with event source?

client/src/components/rooms/transcriptionNotifications/CustomNotifications.tsx
```diff
@@ -0,0 +1,27 @@
...
+  customNotifications?: ActiveNotification[];
+}
+
+export const CustomNotifications = (props: CustomNotificationsProps): JSX.Element => {
+  const { customNotifications } = props;
+  const notificationProps = usePropsFor(NotificationStack);
+  const [activeNotiifications, setActiveNotifications] = React.useState<ActiveNotification[]>([]);
+  useEffect(() => {
+    if (customNotifications) {
+      setActiveNotifications(notificationProps.activeErrorMessages.concat(customNotifications));
```
JamesBurnside, line 18: Does resetting this mess with the dismissal logic? How does the customNotificaiton dismissal get tracked, wondering if new `notificationProps.activeErrorMessages` comes in will they reset the customNotificaiton

client/src/utils/CallAutomationUtils.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    const response = await fetch('/api/summarizeTranscript', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ serverCallId: serverCallId, locale: locale })
+    });
+    console.log('/summarizeTranscript response', response);
```
JamesBurnside, line 144: don't log response, we are logging the summary afterwards. If response needs inspected there's the network tab

client/src/utils/CallAutomationUtils.ts
```diff
@@ -0,0 +1,322 @@
...
+    return;
+  }
+  const remoteParticipantsDisplayInfo = remoteParticipants.map((participant: RemoteParticipantState) => {
+    if ('communicationUserId' in participant.identifier) {
+      return {
+        communicationUserId: participant.identifier.communicationUserId,
+        displayName: participant.displayName
+      };
+    } else {
+      return {};
```
JamesBurnside, line 181: nit: better practice to filter instead of returning an empty object
```
remoteParticipants
  .filter((p) => 'communicationUserId' in participant.identifier)
  .map((p) => ({ communicationUserId: participant.identifier.communicationUserId, displayName: participant.displayName });
```

server/src/app.ts
```diff
@@ -102,7 +104,47 @@ app.use('/api/summarizeTranscript', cors(), summarizeTranscript);
...
+
 const config = getServerConfig();
+/**
+ * route: wss://<host>/
+ * purpose: WebSocket endpoint to receive transcription events
+ *
+ * Don't forget to secure this endpoint in production
+ * https://learn.microsoft.com/en-us/azure/communication-services/how-tos/call-automation/secure-webhook-endpoint?pivots=programming-language-javascript
+ */
+console.log('WebSocket server port:', config.callAutomation?.ServerWebSocketPort);
```
JamesBurnside, line 121: I think this is removed in a later PR but if not remove this

server/src/app.ts
```diff
@@ -102,7 +104,47 @@ app.use('/api/summarizeTranscript', cors(), summarizeTranscript);
...
+    if (
+      ('kind' in messageData && messageData.kind === 'TranscriptionMetadata') ||
+      ('kind' in messageData && messageData.kind === 'TranscriptionData')
+    ) {
+      transcriptionCorrelationId = handleTranscriptionEvent(message, transcriptionCorrelationId);
+    }
+  });
+
+  ws.on('close', () => {
+    console.log('WebSocket closed');
```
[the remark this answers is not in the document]
JamesBurnside: nit: be consistent, sometimes its WebSocket and sometimes its websocket

server/src/routes/callAutomationEvent.ts
```diff
@@ -28,11 +29,14 @@ router.post('/', async function (req, res) {
        * is from the calling SDK and the callConnectionId is mapped to the correlationId from the transcription
        * service. We need to store this mapping so that we can fetch the transcription later.
        */
-      CALLCONNECTION_ID_TO_CORRELATION_ID[callConnectionId] = {
-        serverCallId: serverCallId,
-        correlationId: CALLCONNECTION_ID_TO_CORRELATION_ID[callConnectionId as string]?.correlationId
-      };
       getTranscriptionManager().setCallConnection(callConnectionId, serverCallId);
+      console.log(getTranscriptionManager().getCallConnection(serverCallId));
```
JamesBurnside, line 33: this console log needed?

server/src/routes/callAutomationEvent.ts
```diff
@@ -28,11 +29,14 @@ router.post('/', async function (req, res) {
...
+    } else if (type === 'Microsoft.Communication.TranscriptionStarted') {
+      console.log('/automationEvent received', req.body);
```
JamesBurnside, line 35: these needed, if they're helpful keep them in but be specfic on the event type as both say /automationEvent received

server/src/constants.ts
```diff
@@ -21,12 +21,13 @@ export const VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME =
   'VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT';
 export const VV_COSMOS_DB_CONNECTION_STRING = 'VV_COSMOS_DB_CONNECTION_STRING';
 export const VV_COSMOS_DB_NAME = 'VV_COSMOS_DB_NAME';
-export const VV_AUTO_START_TRANSCRIPTION = 'VV_AUTO_START_TRANSCRIPTION';
+export const VV_TRANSCRIPTION_BEHAVIOR = 'VV_TRANSCRIPTION_BEHAVIOR';
 export const VV_SERVER_WEBSOCKET_URL = 'VV_SERVER_WEBSOCKET_URL';
```
JamesBurnside, line 25: Actually is this needed, is it always the server's own URL? (Could the be queried instead?)

server/src/constants.ts
```diff
@@ -21,12 +21,13 @@ export const VV_POSTCALL_SURVEY_ONEQUESTIONPOLL_SAVE_BUTTON_TEXT_ENV_NAME =
...
 export const VV_SERVER_WEBSOCKET_PORT = 'VV_SERVER_WEBSOCKET_PORT';
 export const VV_SERVER_HTTP_URL = 'VV_SERVER_HTTP_URL';
 export const VV_COGNITIONAPI_KEY = 'VV_COGNITIONAPI_KEY';
```
JamesBurnside, line 28: just a note we should move off these to Azure Credentials authed through managed identity

JamesBurnside (approved):
Approving but please address comments

client/src/components/rooms/transcriptionNotifications/CustomNotifications.tsx
dmceachernmsft: there was a bug here that I fixed later on. this is exactly right, by doing this, this way I had issues seeing the other notifications. (I am also concattenating to the wrong array here but I noticed in a later PR)

client/src/components/rooms/RoomsMeetingExperience.tsx
dmceachernmsft: this does break when deployed but this is retooled later on to get this from the app configuration which fetches it from the server env variables or the `defaultConfig.json` file depending on how contoso is setting up their environment

client/src/components/rooms/transcriptionNotifications/CallSummaryTile.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  {summary ? 'Meeting Summary' : 'Meeting Transcript'}
+                </Text>
+              </Stack>
+
+              <Stack styles={{ root: { padding: '0 1rem' } }}>
+                <Stack
+                  styles={{
+                    root: {
+                      width: '100%',
+                      height: '1px',
```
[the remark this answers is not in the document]
dmceachernmsft: this is just drawing a line to denote a section we dont want it to grow when zooming so should be fine

client/src/components/rooms/transcriptionNotifications/CallSummaryTile.tsx
dmceachernmsft: yes from the server response. this is just a more readable error.

client/src/utils/CallAutomationUtils.ts
dmceachernmsft: this is fixed/ removed later on. we instead have each user update themselves when they join the call.

server/src/app.ts
dmceachernmsft: yes in the end there is no port and is instead integrated into the express server.

server/src/constants.ts
dmceachernmsft: these are needed to set up the env variables when deployed with the blue button and so the server knows how to check for them.

server/src/constants.ts
dmceachernmsft: this is needed in order to hit the language service API in the headers of the request so I don't think we can get around this for now. we will hopefully be able to remove this when CallAutomation releases the transcription and Summary API's as part of their package.

server/src/constants.ts
JamesBurnside: Does it differ to `VV_SERVER_HTTP_URL`?

server/src/routes/callAutomationEvent.ts
dmceachernmsft: These are left in to help the developer see what is happening since these events are streamed in from CallAutomation. I left the template of `automationEvent recieved` so that they will know their endpoint they sent them is working as expected. this is just left in since this is a sample and developers will try to run it locally.

JamesBurnside: why not have `/automationEvent received - transcription started`, or move out of the if else and have `/automationEvent received, type`?

client/src/components/rooms/RoomsMeetingExperience.tsx
dmceachernmsft: I reorganized it alphabetically, I want to keep this simple with using react useStates rather than adding another manager to keep track of these things.

server/src/constants.ts
dmceachernmsft: If you're referring to the websocket URL it differs in the construction of it but the location is the same. needs to be wss://<website> to work whether running locally or deployed.

server/src/routes/callAutomationEvent.ts
dmceachernmsft: So they can see the entire automation event, there is lots of good data in there if they want to do something else with the event.

